PR #33, opened 2019-11-08
<Slider /> and test configurations

allyssonsantos:
This PR implements the Slider component on web and native platforms and some test configurations.

Separated jest setup config for each platform, because we need to specify an "transformIgnorePattern" only in the native environment.

slider-gif


packages/doc/src/components/CodeBlock/CodeBlock.jsx
```diff
@@ -176,10 +175,10 @@ const CodeBlock = ({ children, reactLive, center }) => {
 
   return reactLive ? (
     <MDXContext.Consumer>
-      {scope => (
+      {Scope => (
         <LiveProvider
```
alan-oliv: We changed 'Scope' to capital so we could test the LivePreview
I guess we can change back to lowercase

allyssonsantos: resolved!

PR #62, opened 2019-12-16
[DS-64] Card Plan

ggdaltoso:
**iOS**
Simulator Screen Shot - iPhone 11 - 2019-12-16 at 15 33 58

**Android**
Screenshot_1576522269

**Web**
localhost_8000_components_card_plancard


packages/yoga/src/Card/native/Card/Header.jsx
```diff
@@ -0,0 +1,7 @@
+import styled from 'styled-components';
+
+const Header = styled.View``;
+
+Header.displayName = 'Card.Header';
+
+export default Header;
```
victormath12: Is it necessary to have this Header component in this moment? 'couse it is empty (with no style).

packages/yoga/src/Card/native/Card/Content.jsx
```diff
@@ -0,0 +1,7 @@
+import styled from 'styled-components';
```
victormath12: Is it necessary to have this Content component in this moment? 'couse it is empty (with no style).

PR #68, opened 2020-01-02
[DS-115] Text

victormath12:
Creating Text component (Typography).

image


packages/doc/content/components/components/text/default-native.mdx
```diff
@@ -0,0 +1,72 @@
+import ExpoSnack from 'react-expo-snack';
```
allyssonsantos: You don't need to import `ExpoSnack`
```suggestion

```

packages/yoga/src/Theme/Provider/web/GlobalStyle.jsx
```diff
@@ -0,0 +1,16 @@
+import { createGlobalStyle } from 'styled-components';
+
+const GlobalStyle = createGlobalStyle`
+  ${({
+    theme: {
+      yoga: { baseFontFamily },
+    },
+  }) => `
+    @import url(https://fonts.googleapis.com/css?family=${baseFontFamily}:400,600,700,400i,600i,700i&display=swap);
```
tcK1: This is a really cool solution to load the `font-family`!

BUT, we may want to receive this uri as a parameter that defaults to the google api. We, as a company, should use our CDN to serve these static files, which will be much faster than the Google API.

(We can see this when loading the Yoga documentation and the font flashes when it is loaded [the font was not downloaded yet])

We should also see how should we expose these fonts through the CDN.

packages/doc/src/components/CodeBlock/CodeBlock.jsx
```diff
@@ -281,7 +281,13 @@ const CodeBlock = ({ children, reactLive, center }) => {
       language="jsx"
     >
       {({ className, style, tokens, getLineProps, getTokenProps }) => (
-        <Pre className={className} style={style}>
+        <Pre
+          className={className}
+          style={{
+            ...style,
+            border: '1px solid #DBDBE8',
```
alan-oliv: Why is this inline?

Line 27 declare `Pre` styled-component inheriting gray colors from theme, what you think of adding this style to the component? If you can't find the same gray, use one the look alike from our pallete

victormath12: nice tip! done!

packages/yoga/src/Theme/Provider/web/GlobalStyle.jsx
ggdaltoso: Glad you liked, @tcK1 !

The link you notice us isn't a static file. The Google [Fonts API provides](https://fonts.google.com/about) us an automatically user-agent identification, sending the smallest possible file, caching fonts without compromising user's privacy or security. Also, the cache is made cross-site, making the needs of loading the font just once, with any website and the fonts API is 58% faster than our CDN (measured just for font loading).

The blinking font effect is a UX decision and it could be changed using the [font-display](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display) CSS property which, in this case, is hardcoded as `swap`.

PR #70, opened 2020-01-06
Add state to documentation examples

allyssonsantos:
This PR adds the support to add variables and state to the documentation examples. To do that you need to add the `state` prop to the markdown code block syntax.
Example:

`
```javascript reactLive=true center=true state=true`

Also, change some component examples to add the state on it.

### CODE REVIEW

File: packages/doc/content/components/components/checkbox/switch-web.mdx [hunk 1]
```diff
@@ -5,8 +5,13 @@ Switches are best used for communicating activation (e.g., on/off states), while
 
 ### Usage
 
-```javascript reactLive=true center=true
-<Checkbox.Switch />
+```javascript reactLive=true center=true state=true
+render(() => {
+  const [checked, setChecked] = useState(false);
```
ggdaltoso:
```suggestion
  const [checked, toggleChecked] = useState(false);
```

What do you think of changing `set` to `toggle` on functions that change boolean values?

packages/doc/content/components/components/stepper/stepper-web.mdx
```diff
@@ -4,36 +4,60 @@ Gympass `<Stepper />` component is used to guide the user through determined ste
...
-```javascript reactLive=true center=true
-<Stepper activeStep={0}>
-  <Stepper.Step label="Confirm Booking" />
-  <Stepper.Step label="Class Booked" />
-  <Stepper.Step label="Check-in" />
-</Stepper>
-```
+```javascript reactLive=true center=true state=true
+render(() => {
+  const [activeStep, setActiveStep] = useState(0);
```
ggdaltoso: Could you add a break line between declarations?
💅

allyssonsantos: Done!

packages/doc/content/components/components/checkbox/switch-web.mdx
allyssonsantos: Done!

Issue #73, opened 2020-01-07
[BUG] Plancard has the wrong background color


alan-oliv:
**Describe the bug**
PlanCard appears to use the wrong background color.

**Expected behavior**
The expected background color is 75% of secondary, and not 100%

**Screenshots**
![image](https://user-images.githubusercontent.com/4368481/71898149-58bd8880-3137-11ea-8019-265af025abe0.png)

victormath12:
closed by #76

PR #76, opened 2020-01-09
[DS-144] Card Variant Intensity

victormath12:
In this PR I'm adding a feature to set a variant color and intensity in the `Card` component.

image

**This feature fixes the bug in the Issue #73**


packages/yoga/src/Card/native/Card/Card.jsx
```diff
@@ -19,7 +22,7 @@ const CardStyled = styled.View(
       ${card.padding.left}px;
 
     border-radius: ${card.radii}px;
-    background-color: ${card.backgroundColor};
+    background-color: ${variant ? color[variantIntensity] : white};
```
ggdaltoso:
```suggestion
    background-color: ${variant ? color[variantIntensity] : card.backgroundColor};
```

PR #86, opened 2020-01-31
[DS-93] Embedded expo snacks 

alan-oliv:
### SnackEmbed

This pull request consists of restructuring the `CodeBlock` component for code samples in MDX files.  The restructuring was made in order to add a new component: `SnackEmbed`.

`SnackEmbed` component enables on creating snacks with only a simple code block and with it, there's no more need to create snacks at [https://expo.io/@gympass](url).

It's very important to say that all components documentation were updated to fit the new `SnackEmbed` component and make the process of creating components easier.

Below you can see a simple UML diagram for the `CodeBlock` component new structure.

Untitled Document

_Font imports are not working properly on the snacks (seem to be problematic on web too), we'll open an issue in order to fix that because it's not related to `CodeBlock`._


packages/doc/src/components/CodeBlock/CodeBlock.jsx
```diff
@@ -1,353 +1,97 @@
...
-        colors: { primary: primaryPallete, gray: grayPallete },
-      },
+import React from 'react';
+import { node, bool, oneOf } from 'prop-types';
+import CodeBlockContext from './CodeBlockContext';
+import { ReactLive, PrismHighlight, SnackEmbed } from '.';
+import { importStatement } from './shared/templates';
+import { YogaComponents, YogaIcons, NativeComponents } from './shared/modules';
+
+const CodeBlockComponent = ({ type }) =>
```
ggdaltoso:
```suggestion
const CodeBlockComponent = {
      reactLive: ReactLive,
      highlight: PrismHighlight,
      expo: SnackEmbed,
    };
    
 // using as:
 
 <CodeBlockComponent[type] />
```

What do you think?

packages/doc/src/components/CodeBlock/PrismHighlight/PrismHighlight.jsx
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    div pre {
+        padding: 0 !important;
+    }
+  `}
+`;
+
+const PrismHighlight = ({ code, liveEditor }) => {
+  const highlightCode = !code ? useContext(CodeBlockContext).code : code;
```
ggdaltoso:
```suggestion
  const highlightCode = code || useContext(CodeBlockContext).code;
```

Is it the same?

alan-oliv: Wow! Your suggestion is so much cleaner and beautiful, thanks for that gg <3
I'll refactor this block

packages/doc/src/components/CodeBlock/CodeBlock.jsx
alan-oliv: I like your idea of extracting very much, sir.
This implementation looks different to me though.

I loved your previous idea of making this a gatsby plugin, let's plan this!

alan-oliv:
> An amazing improvement here, @alan-oliv !
> 
> We could extract this SnackEmbed in a separate package in the future.

I like your idea of extracting very much, sir.

I loved your previous idea of making this a gatsby plugin, let's plan this!

packages/doc/src/components/CodeBlock/CodeBlock.jsx
alan-oliv: Loved it!
Discussing with the team, we didn't like of the `<CodeBlockComponent[type] />` directly on JSX, so We adapted, but now is so much simpler without the object literal and find.

Thanks @ggdaltoso

PR #93, opened 2020-02-11
[DS-158] Progress bar

ggdaltoso:
image


packages/doc/content/components/components/progress/progress-web.mdx
```diff
@@ -0,0 +1,27 @@
+### Usage
+
+```javascript reactLive=true
```
alan-oliv: The right prop is `type` but `reactLive` is the default prop, so it's no necessary here

ggdaltoso: updated! (and the CLI template)

PR #94, opened 2020-02-11
[DS-161] Fixing font import in web GlobalStyle

victormath12:
In this PR I'm removing the font import with `@import` and using a third party library `react-google-font-loader` to solve the problem to import google fonts.

### DIFF
--- a/packages/yoga/package.json
+++ b/packages/yoga/package.json
@@ -44,7 +44,8 @@
   },
   "devDependencies": {
     "@testing-library/react": "^9.1.4",
-    "@testing-library/react-native": "^4.1.0"
+    "@testing-library/react-native": "^4.1.0",
+    "styled-components": "^4.4.0"
   },
   "peerDependencies": {
     "styled-components": "^4.4.0"
--- a/packages/yoga/package.json
+++ b/packages/yoga/package.json
@@ -39,7 +39,8 @@
     "@gympass/yoga-tokens": "^0.3.0",
     "@ptomasroos/react-native-multi-slider": "^1.0.0",
     "prop-types": "^15.7.2",
-    "rc-slider": "^8.7.1"
+    "rc-slider": "^8.7.1",
+    "react-google-font-loader": "^1.1.0"
   },
   "devDependencies": {
     "@testing-library/react": "^9.1.4",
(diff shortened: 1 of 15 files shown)


packages/yoga/src/Theme/Provider/web/GlobalStyle.jsx
```diff
@@ -1,16 +1,35 @@
...
+  <>
+    <GoogleFontLoader
+      fonts={[
+        {
+          font: baseFontFamily,
+          weights: [...fontWeights],
+        },
+      ]}
+    />
+    <Global />
```
ggdaltoso:
```suggestion
    <Global fontFamily={baseFontFamily} />
```

What do you think?

packages/yoga/src/Theme/Provider/web/GlobalStyle.jsx
```diff
@@ -1,16 +1,35 @@
...
+  theme: {
+    yoga: { baseFontFamily, fontWeights },
+  },
+}) => (
+  <>
+    <GoogleFontLoader
+      fonts={[
+        {
+          font: baseFontFamily,
+          weights: [...fontWeights],
```
ggdaltoso: If you'll get the `fontWeights` from the theme, you'll need to modify the existent token to provide the missings values (400i,600i,700i).

packages/yoga/src/Theme/Provider/web/expectCssMatches.js
```diff
@@ -3,6 +3,7 @@ const normalizeCSS = rule =>
   rule
     .join('\n')
     .replace(/(?:\r\n|\r|\n)/g, '')
+    .replace(/\/\*.*\*\//g, '')
```
ggdaltoso: Is this regex really necessary?

packages/yoga/package.json
```diff
@@ -39,7 +39,8 @@
     "@gympass/yoga-tokens": "^0.3.0",
     "@ptomasroos/react-native-multi-slider": "^1.0.0",
     "prop-types": "^15.7.2",
-    "rc-slider": "^8.7.1"
+    "rc-slider": "^8.7.1",
+    "react-google-font-loader": "^1.1.0"
```
ggdaltoso: Since this library does not have any tests, do you think we should add some to verify if the tag `link` is created properly?

packages/yoga/src/Theme/Provider/web/GlobalStyle.jsx
victormath12: great, done

packages/yoga/package.json
victormath12: Yes, great point of view, I've done this

PR #121, opened 2020-03-09
Adjusting Dropdown documentation

victormath12:
In this PR I'm adjusting the Dropdown documentation to show examples about `selected` option, as a **plus** I'm adjusting the Navigation to open just the first level of menu.

### DIFF
--- a/packages/doc/content/components/components/button/index.mdx
+++ b/packages/doc/content/components/components/button/index.mdx
@@ -3,5 +3,5 @@ title: 'Button'
 metaTitle: 'Buttons'
 metaDescription: 'This is the list of Buttons'
 linkable: false
-opened: false
+collapsed: false
 ---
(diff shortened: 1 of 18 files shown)


packages/doc/content/components/components/button/index.mdx
```diff
@@ -3,4 +3,5 @@ title: 'Button'
 metaTitle: 'Buttons'
 metaDescription: 'This is the list of Buttons'
 linkable: false
+opened: false
```
alan-oliv: Since the doc menu uses an accordion, what you think of the 'collapsed' nomenclature instead of 'opened'.
On the navigation component, we use 'collapsed' (https://github.com/Gympass/yoga/blob/bebb09b5f47b7946668f14428bc5f300a3192475/packages/doc/src/components/Navigation/Navigation.jsx#L139) 

 It's just a detail though, do you think that makes sense?

victormath12: Nice, already done!

PR #133, opened 2020-04-01
[DS-190] Theme Refactoring

victormath12:
In this PR I'm separating the component's `theme` files to the respective ones, and creating a new CLI step to create the component and the `theme` file automatically.


packages/yoga/src/TextArea/TextArea.theme.js
```diff
@@ -0,0 +1,7 @@
+import Input from '../Input/Input.theme';
+
+const TextArea = tokens => ({
+  ...Input(tokens),
+});
```
ggdaltoso, line 5:
```suggestion
const TextArea = Input;
```

packages/doc/babel.config.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
+  plugins: ['@babel/plugin-proposal-optional-chaining', 'import-glob'],
```
alan-oliv, line 2: shouldn't we use [gatsby-plugin-optional-chaining](https://www.gatsbyjs.org/packages/gatsby-plugin-optional-chaining/) here?

victormath12: Removed!

PR #137, opened 2020-04-29
Create Theme Reader

tcK1:
This PR intends to create a new API to use the styled components yoga theme.

```jsx
// Before:

const MyCoolComponent = styled.div`
  padding: ${(props) => props.theme.yoga.spacing.large}px;
  background-color: ${(props) => props.theme.yoga.colors.primary};
`;

// After:
import { theme } from '@gympass/yoga';

const { spacing, colors } = theme;

const MyCoolComponent = styled.div`
  padding: ${spacing.large}px;
  background-color: ${colors.primary};
`;
``` 

- **Important**:

  [GoogleChrome/proxy-polyfill](https://github.com/GoogleChrome/proxy-polyfill) was added to make it compatible with IE.

---

The cool next step would be to create a custom babel plugin to replace the values.


alan-oliv:
Hi @tcK1, I'm extremely grateful for your contribution! Thanks! 🤩

I'd like to open a discussion here about this enhancement so I can hear you and the team's opinion, about it.

Although I did find your `Proxy` solution very elegant used in order to simplify the code writing on a `styled-component` I'm a lil bit concerned about its use on a complex component, thus your solution executes a callback on each line that needs a prop from theme

Take this [Button](https://github.com/Gympass/yoga/blob/master/packages/yoga/src/Button/web/StyledButton.jsx) for instance, each line use a property from theme but it only calls one callback, instead of 27 callbacks using the themeReader 

So, what everybody thinks? how can we solve that?

PS: Nice job on the GoogleChrome/proxy-polyfill

tcK1:
Hey @alan-oliv! Calling a function multiple times could be troublesome. One thing to remember is that this approach does not invalidate the other.

I created this "helper" with the intuit to "help" developers use the yoga theme inside their on components, which is much simpler then the components inside our system. Here's an [example](https://github.com/Gympass/albus/blob/56623772f83c01af04035b473f1a8872d7d01e5e/components/ClassOver/styles.js).

If we wanted to use the helper on our components, we also could do that with a single function call. Here's how:

Before:
```js
const Button = styled.button`
  ${({
    full,
    small,
    inverted,
    theme: {
      yoga: {
        baseFont,
        components: { button },
      },
    },
  }) => `
    /* css goes here... also, this indentation is so weird haha. */
    `}
`;
```

After:
```js
const Button = styled.button`
  ${props => {
    const { full, small, inverted } = props;

    const {
      baseFont,
      components: { button },
    } = theme(props); // The single function call.

    return css`
      /* css goes here... */
    `;
  }
`;
```

One thing that could be possible, is to create a "cache" for the proxy. I will try to come up with something.

tcK1:
I've committed a change that adds cache to the `Proxy` generator. Here are the results:

The first spec was:

```js
  const obj = {
      theme: {
        yoga: {
          breakpoints: {
            xs: {
              margin: 'value',
            },
          },
        },
      },
    };

    theme(obj);
    theme.breakpoints(obj);
    theme.breakpoints.xs(obj);
    theme.breakpoints.xs.margin(obj);
```

Before:
```sh
  console.log packages/yoga/src/Theme/helpers/themeReader/index.js:17
    Created Proxy for [ 'theme', 'yoga' ]

  console.log packages/yoga/src/Theme/helpers/themeReader/index.js:17
    Created Proxy for [ 'theme', 'yoga', 'breakpoints' ]

  console.log packages/yoga/src/Theme/helpers/themeReader/index.js:17
    Created Proxy for [ 'theme', 'yoga', 'breakpoints' ]

  console.log packages/yoga/src/Theme/helpers/themeReader/index.js:17
    Created Proxy for [ 'theme', 'yoga', 'breakpoints', 'xs' ]

  console.log packages/yoga/src/Theme/helpers/themeReader/index.js:17
    Created Proxy for [ 'theme', 'yoga', 'breakpoints' ]

  console.log packages/yoga/src/Theme/helpers/themeReader/index.js:17
    Created Proxy for [ 'theme', 'yoga', 'breakpoints', 'xs' ]

  console.log packages/yoga/src/Theme/helpers/themeReader/index.js:17
    Created Proxy for [ 'theme', 'yoga', 'breakpoints', 'xs', 'margin' ]
```
After:
```sh
  console.log packages/yoga/src/Theme/helpers/themeReader/index.js:17
    Created Proxy for [ 'theme', 'yoga' ]

  console.log packages/yoga/src/Theme/helpers/themeReader/index.js:17
    Created Proxy for [ 'theme', 'yoga', 'breakpoints' ]

  console.log packages/yoga/src/Theme/helpers/themeReader/index.js:17
    Created Proxy for [ 'theme', 'yoga', 'breakpoints', 'xs' ]

  console.log packages/yoga/src/Theme/helpers/themeReader/index.js:17
    Created Proxy for [ 'theme', 'yoga', 'breakpoints', 'xs', 'margin' ]
```

The second result is the behavior that we should expect.

The second spec was:

```js
  const obj = {
      theme: {
        yoga: {
          breakpoints: {
            xs: {
              margin: 'value',
            },
          },
        },
      },
    };

    theme.breakpoints.xs.margin(obj);
    theme.breakpoints.xs.margin(obj);
    theme.breakpoints.xs.margin(obj);
    theme.breakpoints.xs.margin(obj);
    theme.breakpoints.xs.margin(obj);
    theme.breakpoints.xs.margin(obj);
    theme.breakpoints.xs.margin(obj);
    theme.breakpoints.xs.margin(obj);
    theme.breakpoints.xs.margin(obj);
    theme.breakpoints.xs.margin(obj);
```

Before:

```sh
  console.log packages/yoga/src/Theme/helpers/themeReader/index.js:17
    Created Proxy for [ 'theme', 'yoga' ]

  console.log packages/yoga/src/Theme/helpers/themeReader/index.js:17
    Created Proxy for [ 'theme', 'yoga', 'breakpoints' ]

  console.log packages/yoga/src/Theme/helpers/themeReader/index.js:17
    Created Proxy for [ 'theme', 'yoga', 'breakpoints', 'xs' ]

  console.log packages/yoga/src/Theme/helpers/themeReader/index.js:17
    Created Proxy for [ 'theme', 'yoga', 'breakpoints', 'xs', 'margin' ]

 # Repeat 10 times.
```

After:

```sh
  console.log packages/yoga/src/Theme/helpers/themeReader/index.js:17
    Created Proxy for [ 'theme', 'yoga' ]

  console.log packages/yoga/src/Theme/helpers/themeReader/index.js:17
    Created Proxy for [ 'theme', 'yoga', 'breakpoints' ]

  console.log packages/yoga/src/Theme/helpers/themeReader/index.js:17
    Created Proxy for [ 'theme', 'yoga', 'breakpoints', 'xs' ]

  console.log packages/yoga/src/Theme/helpers/themeReader/index.js:17
    Created Proxy for [ 'theme', 'yoga', 'breakpoints', 'xs', 'margin' ]

  # Only create the Proxy a single time.
```

---

Since the theme `Proxy` is application wide, if multiple components use the same values (`theme.breakpoints.xs.margin` in this example), it will only create the `Proxy` a single time. This is also valid if used inside the yoga components.

I hope that this can add some ideas to the discussion.

ggdaltoso:
Thanks for your suggestion, @tcK1, and for the clever implementation as well.

Taking your example, we could simplify it by doing this:
```diff
export const LottieWrapper = styled.div`
  width: 208px;
  height: 162px;
- margin:
-   0
-   0
-   ${({ theme: { yoga } }) => yoga.spacing.small}px
-   0;
- @media (min-width: 1024px) {
-   width: 296px;
-   height: 231px;
-   margin:
-     0
-     ${({ theme: { yoga } }) => yoga.spacing.xxlarge}px
-     0
-     0;
- }

// I would suggest you to only put styles which depends on a theme inside the callback
+  margin: 0;
+
+  @media (min-width: 1024px) {
+    width: 296px;
+    height: 231px;
+  }
+
+  ${({
+    theme: {
+      yoga: {
+        spacing
+      }
+    }
+  }) => `
+  margin-bottom: ${spacing.small}px;
+
+  @media (min-width: 1024px) {
+    width: 296px;
+    height: 231px;
+    margin-right: ${spacing.xxlarge}px;
+  }
`;
```

Personally, I think it is easy to read in this way.

Taking another component from your example, to ilustrate what @alan-oliv is talking about, we have something like this:

```jsx
// The original
export const Title = styled(Text.H1)`
  margin-bottom: ${({ theme: { yoga } }) => yoga.spacing.xsmall}px; // 1º callback
  @media (min-width: 1024px) {
    margin-bottom: ${({ theme: { yoga } }) => yoga.spacing.xxlarge}px; // 2º callback
  }
`;

// The Yoga way (as we do)
export const Title = styled(Text.H1)`
  /* 1º callback */
  ${({ theme: { yoga: { spacing } } }) => ` 
    margin-bottom: ${spacing.xsmall}px;
    
    @media (min-width: 1024px) {
      margin-bottom: ${spacing.xxlarge}px;
    }
  `}
`;

// Using your solution (without destructuring)
export const Title = styled(Text.H1)`
  /* 1º callback */
  margin-bottom: ${theme.spacing.xsmall}px; 
  
  /* 2º callback */
  @media (min-width: 1024px) {
    margin-bottom: ${theme.spacing.xxlarge}px; 
  }
`;

// Using your solution (with destructuring)
export const Title = styled(Text.H1)`
  /* 1º callback */
 ${props => { 
  const {
    spacing,
  } = theme(props); /* 2º callback */
  margin-bottom: ${spacing.xsmall}px;

  @media (min-width: 1024px) {
    margin-bottom: ${spacing.xxlarge}px;
  }
`;
```

IMO, the way we do here at Yoga is easier to read and it scales very well as we can see by the example above, and despite of any CSS property, we could have just one callback.

I want to know more about why you need this. There is too many ``` ${({ theme: { yoga } }) => yoga.*} ``` in your code?

tcK1:
@ggdaltoso 
> I want to know more about why you need this. There is too many ${({ theme: { yoga } }) => yoga.*} in your code?

Yeah. Not only that, we could do something like we do inside the Yoga components (single function call) but we lose syntax highlight and `eslint`+`prettier`

tcK1:
@ggdaltoso @alan-oliv 

I decide to run a little test. Compare the render time of the ways we could write the component, these were the results:

```sh
  console.time packages/yoga/src/Theme/helpers/themeReader/web/web.test.jsx:259
    Render StyledButtonWithoutHelper: 18ms

  console.time packages/yoga/src/Theme/helpers/themeReader/web/web.test.jsx:267
    Render StyledButtonWithHelperSingleFunction: 9ms

  console.time packages/yoga/src/Theme/helpers/themeReader/web/web.test.jsx:275
    Render StyledButtonWithHelperMultipleFunctions: 9ms
```

And this was the spec I ran: https://pastebin.com/5LVVe3Qr (too big to paste here hahah)

Also, while doing this I found an error on the `Button` component. There's a `:` missing in [this line](https://github.com/Gympass/yoga/blob/master/packages/yoga/src/Button/web/StyledButton.jsx#L55). The `eslint`+`prettier` found the error for me (another vouch for in my previous comment).

alan-oliv:
@tcK1 you had me with the cache implementation and the performance tests 👏🏻

It took away my concerns, so I have no reasons not to approve this PR if will help people out on using Yoga

I had a lot a fun with all the tests and the discussions, thank you for your contribution and effort

tcK1:
Another solution, could be using css variables instead (we would need the [ponyfill](https://www.npmjs.com/package/css-vars-ponyfill)). I don't know what would be better, but I kinda prefer the `Proxy` since it give more flexibility (and the user can't overwrite it).

allyssonsantos:
@all-contributors please add @tcK1 for code, ideas, doc

PR #142, opened 2020-05-18
Change Corporate theme colors

tsuzukayama:
We have a color pallete in Clients/Corp that doesn't match the one provided. So we need to change it.

* Add color `buenosAires`;
* Change primary color from Corporate theme to `buenosAires`;
* Change secondary color from Corporate theme to `newYork`;

### DIFF
--- a/packages/yoga/src/Theme/themes/Corporate.js
+++ b/packages/yoga/src/Theme/themes/Corporate.js
@@ -2,8 +2,8 @@ import BaseTheme from './BaseTheme';
 
 const Corporate = tokens => {
   const colors = {
-    primary: tokens.colors.newYork,
-    secondary: tokens.colors.paris,
+    primary: tokens.colors.buenosAires,
+    secondary: tokens.colors.saoPaulo,
   };
 
   return { colors, ...BaseTheme(colors) };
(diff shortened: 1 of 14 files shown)


ggdaltoso:
Hi, @tsuzukayama

Could you, please, share with us where do these values come from?

Actually, we have [this](https://www.figma.com/file/uixBrcFa8f9VSmwCE9zuPj/Design-System-1.0?node-id=3988%3A0):
Clients:
 - **Primary**: SaoPaulo
 - **Secondary**: NewYork

tsuzukayama:
Hi @ggdaltoso, Mariano Dantas And Marcelus Fernandes (PD's) created this new color for Clients. You could talk to them to check it.

packages/yoga/src/Theme/themes/Corporate.js
```diff
@@ -2,8 +2,8 @@ import BaseTheme from './BaseTheme';
 
 const Corporate = tokens => {
   const colors = {
-    primary: tokens.colors.newYork,
-    secondary: tokens.colors.paris,
+    primary: tokens.colors.buenosAires,
+    secondary: tokens.colors.newYork,
```
alan-oliv, line 6: @tsuzukayama talking with the product designers, we figured that the secondary color pallete for corp is actually `saoPaulo`

tsuzukayama: @alan-oliv I updated the `buenosAires` and `saoPaulo` colors. Do you mind taking a look? There were some snapshot updates too, since `saoPaulo` is used on other places.

PR #145, opened 2020-06-16
[CON-2318] Add Slider Tooltip option: `alwaysShow`

tcK1:
We want to always show the tooltip, even if the user have not pressed the marker.

desk
gifsmallersmaller


allyssonsantos:
Hi @tcK1,
Why do you need to the tooltip be always visible? Do you have a case to show us, for example?

tcK1:
Sure thing, @allyssonsantos!

When the user opens the gym search filter, we want to be able to have the current selected tooltip displayed, instead of an open space.

Currently we have this behavior:

image

@alan-oliv FYI.

tcK1:
While speaking with @alan-oliv, we understood that this is not the correct approach to solve our problem.

PR #168, opened 2020-10-19
docs(native-font): add documentation for native font in getting started

caioalexandrebr:
Hi guys, how are you? :rocket:

It is a pleasure to contribute to this incredible project. I found an item about the documentation where I went through difficulty.

The native font import documentation was in the `Text` component. The components are listed in alphabetical order, so the `Text` component becomes one of the last items, making it difficult to access font import documentation.

### DIFF
--- a/packages/doc/content/components/getting-started/default-native.mdx
+++ b/packages/doc/content/components/getting-started/default-native.mdx
@@ -1,15 +1,3 @@
-### Installing
-
-In order to install our design-system just run:
-
-```bash type=highlight
-yarn add @gympass/yoga
-
-// or
-
-npm install @gympass/yoga
-```
-
 ### Adding Gympass font in your project
 
 Both iOS and Android come with their own set of platform fonts but all Yoga
@@ -53,18 +41,4 @@ and run the command bellow:
 
 ```javascript type=highlight
 react-native link
-```
-
-### Usage
-
-An important point in using it is that your whole application must be wrapped in our `ThemeProvider` component:
-
-```javascript type=highlight
-import { ThemeProvider, Button } from '@gympass/yoga';
-
-const App = () => (
-  <ThemeProvider>
-    <Button>Find an activity</Button>
-  </ThemeProvider>
-);
 ```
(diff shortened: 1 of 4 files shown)


packages/doc/content/components/getting-started/default-native.mdx
```diff
@@ -0,0 +1,70 @@
...
+
+In order to install our design-system just run:
+
+```bash type=highlight
+yarn add @gympass/yoga
+
+// or
+
+npm install @gympass/yoga
+```
```
ggdaltoso, line 11: I think we could place these lines in `packages/doc/content/components/getting-started/index.mdx` since the installation step is the same for both (web and native).

What do you think?

caioalexandrebr: Hi @ggdaltoso ,

Thanks for the comment, it makes perfect sense, i made the change, what do you think? :rocket:

packages/doc/content/components/getting-started/default-web.mdx
```diff
@@ -0,0 +1,28 @@
...
+      <link
+        rel="stylesheet"
+        href={`https://fonts.googleapis.com/css?family=${family}:${weight.join()}`}
+      />
+    )}
+  </>
+);
+
+export default withTheme(FontLoader);
+```
```
allyssonsantos, line 28: Hi! @caioalexandrebr Thanks for your contribution!

In this case, The `FontLoader` component must be imported from `@gympass/yoga`. [Yoga already exports a component](https://github.com/Gympass/yoga/blob/master/packages/yoga/src/index.js#L21) to do this job, there's no need to add a new external library, like "react-google-font-loader".

So, In my opinion, I think that the example must be like the `ThemeProvider` component example.

allyssonsantos (changes requested):

packages/doc/content/components/getting-started/default-web.mdx
caioalexandrebr: Hi @allyssonsantos !! :satisfied:

Very good your comment, I made the changes, now I think it is meeting expectations, what do you think?

allyssonsantos (approved):

ggdaltoso (approved):

ggdaltoso:
@all-contributors please add @caioalexandrebr for doc

PR #169, opened 2020-10-21
New icons!

ggdaltoso:
image

This PR also includes the alpha and beta release on ci 🙏


ggdaltoso (commented):
LGTM

allyssonsantos (approved):

scripts/publish.sh
```diff
@@ -0,0 +1,13 @@
...
+​
+if [ $BRANCH == 'master' ];
+then
+  yarn lerna publish --yes
+elif [ $BRANCH == 'alpha' ] || [ $BRANCH == 'beta' ];
+then
+  yarn lerna publish --conventional-prerelease --yes --preid $BRANCH
+else
+  echo "Nothing to publish on branch $BRANCH..."
+fi
```
alan-oliv, line 13: Great implementation but I'd rather do this inside the CI pipeline.
IMO CI is responsible and not a shell script since for `lerna publish` runs properly I need to be logged in npm at my local machine, and so, I'd move this from the application to the CI.

If you guys agree, we can do this as a improvement (since this is already very nice)

alan-oliv (approved):
This is awesome. 
I learned a lot with this PR, thanks for that @ggdaltoso and @allyssonsantos

PR #177, opened 2020-10-21
Add phone icon

tcK1:
image

- Add `Phone` SVG icon.


ggdaltoso:
This PR #169 contains a lot of new icons and one of them is for smartphones. We didn't publish yet because we're trying to make a prerelease of Yoga.

Does this icon work for you?

coelhucas:
> This PR #169 contains a lot of new icons and one of them is for smartphones. We didn't publish yet because we're trying to make a prerelease of Yoga.
> 
> Does the Smartphone icon work for you?

It's not the exactly same that we'll use, but I don't know if it's a problem. Do you have a time estimate to this being published?

tcK1:
@ggdaltoso waiting for the #169 is probably the best approach. 😄 

I don't know where the designers get the icons from Figma, but if we have this different icon we should probably revisit it (the Figma library).

I will close this PR in favor of the #169.

ggdaltoso:
@tcK1 We've just published an alpha Yoga version, [5.1.1-alpha.0](https://www.npmjs.com/package/@gympass/yoga/v/5.1.1-alpha.0), containing all icons. 

Could you check if it works for you?

PR #197, opened 2020-12-09
hide maxLength on Input

thayllachristine:
I added `hideMaxLength` prop in `<Helper/>`, because we need to hide the maxLegnth count into `<Input/>`. It’s unnecessary to presents this info to user in SingUp scenario (when user insert your national document, specifically in second step on the flow).


allyssonsantos (approved):

ggdaltoso (approved):
Thanks for your contribution @thayllachristine !!!

allyssonsantos:
@all-contributors please add @thayllachristine for code

PR #198, opened 2020-12-09
@gympass/yoga-helpers: new package! 🎉

ggdaltoso:
Starting this new package with `media` module!

It helps us to write responsible code with `styled-components`!!!

``` javascript
import { media } from '@gympass/yoga-helper';
​
const MyResponsiveButton = styled.button`
  ${media[breakpoint]``}
  // generates: @media (min-width: breakpoint)
  // Ex.: media.md`` or media['xxs']``
​
  ${media.min(breakpoint)``}
  // generates: @media (min-width: breakpoint)
  // Ex.: media.min('md')``

  ${media.max(breakpoint)``}
  // generates: @media (max-width: breakpoint)
  // Ex.: media.max('xl')``
​
  ${media.between(minBreakpoint, maxBreakpoint)``}
  // generates: @media (min-width: minBreakpoint) and (max-width: maxBreakpoint)
  // Ex.: media.between('md', 'l')``
​
  // We can also use .not (Like jest expect matchers)
  ${media.not[breakpoint]``}
  ${media.not.min(breakpoint)``}
  // generates: @media not all and (min-width: breakpoint)
  // Ex.: media.not.md`` or media.not['xxs']``

  ${media.not.max(breakpoint)``}
  // generates: @media not all and (max-width: breakpoint)
  // Ex.: media.not.max('xxl')``

  ${media.not.between(minBreakpoint, maxBreakpoint)``}
  // @media not all and (min-width: breakpoint) and (max-width: breakpoint)
  // Ex.: media.not.between('sm', 'mdx')``
`;
```

And also hiding!

```javascript
import { media } from '@gympass/yoga-helper';

const HideMe = styled.div`
  ${media.hide.xxs}
  // will hide for, and only for, xxs

  ${media.hide['md-start']}
  // you can define when it will start hiding
`;

const NotHideMe = styled.div`
  ${media.not.hide.sm} // will hide for everyone besides sm
`;
```


packages/helpers/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "homepage": "https://gympass.github.io/yoga/",
+  "license": "MIT",
+  "main": "src/index.js",
+  "directories": {
+    "lib": "lib",
+    "test": "__tests__"
+  },
+  "files": [
+    "lib"
+  ],
```
allyssonsantos, line 21: Those lines are unnecessary.
```suggestion
```

allyssonsantos (changes requested):

packages/helpers/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "publishConfig": {
+    "access": "public"
+  },
```
allyssonsantos, line 24:
```suggestion
  "publishConfig": {
    "access": "public",
    "directory": "dist"
  },
```

We need to add the `directory` field to tell npm to publish the right bundle

allyssonsantos (changes requested):

packages/doc/content/components/helpers/media.mdx
```diff
@@ -0,0 +1,138 @@
...
+  ${media.hide['md-start']}
+`;
+
+const NotHideMe = styled.div`
+  // will hide for everyone besides sm
+  ${media.not.hide.sm}
+`;
+```
+
+### Example
```
tcK1, line 84: Do we need both `Usage` and `Example` sections? In my opinion, we could have only the `Example` section but name it as `Usage` 😛 

What do you guys think?

packages/helpers/src/hide.js
```diff
@@ -0,0 +1,95 @@
+import tokens from '@gympass/yoga-tokens';
+import { css } from 'styled-components';
+import { not } from './media';
+
+const {
+  breakpoints: { xxs, xs, sm, md, lg, xl, xxl, xxxl },
+} = tokens;
+
+const hide = isNot => ({
```
tcK1, line 9: Could we make it dynamic? If we change the breakpoints, we will need to update the code here too, right?

```js
const breakpointsAsKeys = Object.keys(breakpoints);

const generateMedias = isNot => breakpointsAsKeys.reduce((acc, breakpoint, index) => {
  const breakpointValue = breakpoints[breakpointsAsKeys[index]];

  if (index === 0) {
    acc[breakpoint] = css`
      @media ${not(isNot)} (max-width: ${breakpointValue.width}px) {
        display: none !important;
      }
    `;
    acc[`${breakpoint}-start`] = css`
      @media ${not(isNot)} (min-width: ${breakpointValue.width}px) {
        display: none !important;
      }
    `;

    return acc;
  }

  if (index === breakpointsAsKeys.length) {
    acc[breakpoint] = css`
      @media ${not(isNot)} (min-width: ${breakpointValue.width}px) {
        display: none !important;
      }
    `;

    return acc;
  }

  const nextBreakpointValue = breakpoints[breakpointsAsKeys[index + 1]];

  acc[breakpoint] = css`
    @media ${not(isNot)}
      (min-width: ${breakpointValue.width}px) and
      (max-width: ${nextBreakpointValue.width - 1}px)
    {
      display: none !important;
    }
  `;

  acc[`${breakpoint}-start`] = css`
    @media ${not(isNot)} (min-width: ${breakpointValue.width}px) {
      display: none !important;
    }
  `;

  return acc;
}, {});

const normalMedias = generateMedias(false);
const oppositeMedias = generateMedias(true); // We may need a better name :P

const hide = isNot => isNot ? oppositeMedias : normalMedias;
```

I'm not sure if this code works, it's just to give an idea.

packages/helpers/src/media.js
```diff
@@ -0,0 +1,53 @@
...
+export const not = isNot => (isNot ? 'not all and ' : '');
+
+const getRange = (width, range) => {
+  if (Array.isArray(width)) {
+    const [min, max] = width;
+
+    const {
+      [min]: { width: minBreakpoint },
+      [max]: { width: maxBreakpoint },
+    } = breakpoints;
```
tcK1, line 21: Do we need to `try/catch` this block to return a better error? Invalid parameters, for example.

```js
media.between('sm', 'lgg') // Misspelling
```

packages/helpers/src/media.js
```diff
@@ -0,0 +1,53 @@
...
+    ${css(...style)}
+  }
+`;
+
+Object.keys(breakpoints).reduce((acc, label) => {
+  acc[label] = matcher(label);
+  acc.not[label] = matcher(label, true);
+
+  return acc;
+}, media);
```
tcK1, line 42: This is not being used, right? 🤔

packages/doc/content/components/helpers/media.mdx
ggdaltoso: I did what you suggested and it looks strange to me. I miss this section break between _what is available for you to use_ and _how to use it_.

In `Usage` I wanted to show the `media` API whereas the `Example` has a real instance of what could be used in a real-world application. 

It would be better to rename `Usage` to `API` and keep `Installing`, `API`, and `Usage`or just `Installing` and `Usage`?

packages/helpers/package.json
ggdaltoso: Done!

packages/helpers/src/media.js
ggdaltoso: This is the code responsible for adding the defaults: 
 `media.xxs`, `media.xs`, `media.sm`, `media.md`, `media.lg`, `media.xl`, `media.xxl`, `media.xxxl`, `media.not.xxs`, `media.not.xs`, `media.not.sm`, `media.not.md`, `media.not.lg`, `media.not.xl`, `media.not.xxl`, and `media.not.xxxl`

packages/doc/content/components/helpers/media.mdx
tcK1: I like the ideia of having a separate section for the API and how to use it. My comment was based on what `styled-components` [does on its documentation](https://styled-components.com/docs/api#css).

If we want to showcase the API "possibilities" we could add a simpler segment with only the Object options, not necessarily inside a styled-component. For example:

```js
media[breakpoint]``
// Ex.: media.md`` or media['md-start']``

media.min(breakpoint)``
media.max(breakpoint)``
// Ex.: media.min('md')``

media.between(minBreakpoint, maxBreakpoint)``
// Ex.: media.between('md', 'l')``


// We can also use .not (Like jest expect matchers)
media.not[breakpoint]``

media.not.min(breakpoint)``
media.not.max(breakpoint)``

media.not.between(minBreakpoint, maxBreakpoint)``
```

So we would keep: `Installing`, `API`, and `Usage`. Would that have all the information the developer needs to use the helper?

packages/doc/content/components/helpers/media.mdx
```diff
@@ -0,0 +1,138 @@
...
+  ${media.max('xl')``}
+​
+  // for a device range
+  // @media (min-width: breakpoint) and (max-width: breakpoint)
+  ${media.between('md', 'l')``}
+​
+  // jest like
+  // @media not all and (min-width: breakpoint)
+  ${media.not.md``}
+
```
evilamaior, line 52: what do you think about including this `media.not.min`?

packages/helpers/src/media.js
tcK1: Hmmm, got it! I did not see the `media` as the initial value. 🤦🏼 

Maybe we could use another approach to make this more "obvious":

```suggestion
// Add default values.
Object.keys(breakpoints).forEach((label) => {
  media[label] = matcher(label);
  media.not[label] = matcher(label, true);
});
```

packages/helpers/src/media.js
ggdaltoso: Done!

packages/helpers/src/hide.js
ggdaltoso: Done!

packages/doc/content/components/helpers/media.mdx
ggdaltoso: Could you review it again, please?

packages/doc/content/components/helpers/media.mdx
ggdaltoso: Much better!

tcK1 (approved):
Nice! 🚀

evilamaior (approved):

allyssonsantos (approved):

thayllachristine (approved):

PR #212, opened 2021-02-09
Add IconWrapper abstraction

coelhucas:
### About

This adds the new IconWrapper abstraction which receives token names as string and handle their values without overhead. Original API suggested by @ggdaltoso [here](https://gympass.slack.com/archives/CM6G11SF8/p1612812228023600?thread_ts=1612810813.020600&cid=CM6G11SF8).

This abstraction should help us with the trade off where instead of having to generate a new styled component to apply our theme every time we change its fill or stroke, the Wrapper will automagically resolve the received token names to its respective values.

### Images
image

<sup>You can run it locally to check the current documentation.</sup>

- [x] Web tested
- [x] iOS tested
- [x] Android tested
- [ ] Expo example tested


coelhucas:
I wasn't able to run it in expo. But I tried other components such as List and Dropdown and I received the same error:

tcK1:
What if the icon itself already had this logic? Removing the need of this "wrapper":

```jsx
import { Close } from '@gympass/yoga-icons';

/* ... */

<Close fill="primary" height="small" />
```

I didn't like needing to import 2 things to use an icon (`svg`), just my opinion. 🤔

geovanasilva (approved):

ggdaltoso:
@tcK1 

If we change the icons package to manage which colors all icons should have, we might be mixing things

Making the `IconWrapper` will not only work with `@gympass/yoga-icons` but all SVGs. The responsibility of taking the correct color or setting the `width` and `height` properly against our theme is from Yoga and I think this component is doing it great in this way.

coelhucas:
Tossing my two cents, even not being the best to import Wrapper + Icons individually, I think it's a tradeoff much better than what we were previously doing, which was more busywork between the style file and the component just to use a custom color.

tcK1:
@ggdaltoso @coelhucas You guys are right!

What if we had shortcuts for icons that already exist on the `yoga-icons` package?

```js
//           \/\/\/\/\/\/  
<IconWrapper icon="Close" fill="primary" height="small" />
```
> `string` instead of the `svg`

What do you guys think?

allyssonsantos:
> What if the icon itself already had this logic? Removing the need of this "wrapper":
> 
> ```js-jsx
> import { Close } from '@gympass/yoga-icons';
> 
> /* ... */
> 
> <Close fill="primary" height="small" />
> ```
> 
> I didn't like needing to import 2 things to use an icon (`svg`), just my opinion. 🤔

@tcK1 

We are "generating" the icons via a babel plugin configuration, as you can see [here](https://github.com/Gympass/yoga/blob/alpha/packages/icons/babel.config.js), so, to implement this suggestion we will need to write our own plugin to handle this case, we can discuss about it in the future, but, I think that's the most simple solution to handle this problem.

> @ggdaltoso @coelhucas You guys are right!
> 
> What if we had shortcuts for icons that already exist on the `yoga-icons` package?
> 
> ```js
> //           \/\/\/\/\/\/  
> <IconWrapper icon="Close" fill="primary" height="small" />
> ```
> 
> > `string` instead of the `svg`
> 
> What do you guys think?

To implement this approach we will need the same case as above, and I think that we are going to face a new one too.
We can face some tree-shaking problems with this approach, to handle it we will need to import all the available icons inside `yoga-icons` and find the one that we receive via prop. So, it will be ~imposible~ hard to "tree-shake" it.

tcK1:
> We are "generating" the icons via a babel plugin configuration, as you can see [here](https://github.com/Gympass/yoga/blob/alpha/packages/icons/babel.config.js?rgh-link-date=2021-02-09T19%3A41%3A22Z), so, to implement this suggestion we will need to write our own plugin to handle this case, we can discuss about it in the future, but, I think that's the most simple solution to handle this problem.

Yeah, `svgr` is awesome. We could solve this with JS and React, but that would **probably** also have tree-shaking problems. 😢 

```js
// Pseudo

import * as Icons from './svgs'

export default Icons.reduce(/* Transform each Icon to the magic IconWrapper component */, {});
``` 

Although, as @ggdaltoso said, this is more philosophical question: "Should the icons know about the theme?"

If I understood it correctly, no, it shouldn't.

---

Considering these topics, I think that the double import notation is the best solution after all. 😂

coelhucas:
So... do we keep it as is?

alycecristines (approved):

packages/yoga/src/IconWrapper/IconWrapper.jsx
```diff
@@ -0,0 +1,52 @@
...
+  'huge',
+  'xhuge',
+];
+
+IconWrapper.propTypes = {
+  as: elementType.isRequired,
+  fill: string,
+  stroke: string,
+  width: oneOf(commonSizes),
+  height: oneOf(commonSizes),
```
tcK1, line 42: If we also wan't to be able to use "custom" values, we should reflect it at the `propTypes`, don't we?

`allColors || String` and `allSizes || Number`.

packages/yoga/src/IconWrapper/IconWrapper.jsx
```diff
@@ -0,0 +1,52 @@
...
+};
+
+IconWrapper.defaultProps = {
```
tcK1, line 45: Maybe we don't need `defaultProps`, since the `svg` itself already has it. 🤔

packages/yoga/src/IconWrapper/IconWrapper.jsx
```diff
@@ -0,0 +1,52 @@
...
+  width,
+  height,
+  fill,
+  stroke,
+  theme,
+  ...props
+}) => (
+  <Component
+    width={theme.yoga.spacing[width] || width}
+    height={theme.yoga.spacing[height] || height}
```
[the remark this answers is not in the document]
ggdaltoso: `spacing` is an array with [some aliases](https://github.com/Gympass/yoga/blob/master/packages/tokens/src/global/spacing.js) mapping its indexes.

```javascript
theme.yoga.spacing[5] === theme.yoga.spacing.small // true
```

So, if you want `width`, for example, to be `5px` you just pass it as `"5px"`. If you pass it as `5` or `small`, it would set the SVG width as 16px (current `spacing[5]`)

This prop should be `string` or `number`. Right?

packages/yoga/src/IconWrapper/IconWrapper.jsx
coelhucas: Why would we remove it? We're not only sending directly values to svg and to remove `defaultProps` we would need to make fill, stroke and etc required properties rather than letting a common default.

dmscn (approved):

ggdaltoso (approved):

packages/yoga/src/IconWrapper/IconWrapper.jsx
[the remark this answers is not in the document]
tcK1: Yeah, it's just that plain numbers are also valid for these attributes, we are "reserving" `0` to `11`.

But if that's clear, I see no problem!

packages/yoga/src/IconWrapper/IconWrapper.jsx
tcK1: Hmmmm, would this work?

```js
oneOf([...commonSizes, instanceOf(Number)])
```

Pretty sure it won't, but we can try. 🤣

Another option would be this (as @ggdaltoso [said](https://github.com/Gympass/yoga/pull/212#discussion_r573854431)):

```js
oneOfType([string, number])
```

packages/yoga/src/IconWrapper/IconWrapper.jsx
tcK1: Ah, you are right, we need default props if the prop is not required. Sorry. 🤦🏼‍♂️ 

I think what I meant is: Why can't all these default values be `undefined` since the `svg` itself already has the default values.

packages/yoga/src/IconWrapper/IconWrapper.jsx
coelhucas: Actually, the first suggestion didn't triggered any console error, warning or crash. Nice to know, never thought of it 🤔

allyssonsantos (approved):

packages/doc/content/components/components/icon/index.mdx
```diff
@@ -0,0 +1,33 @@
...
+
+
+# Icon
+
+This icon works as a wrapper for our icons abstracting color, size and stroke usage, which are all received as strings
+representing a spacing token name.
+
+* Refer to [Icons](icons) to check already existing icons which can be used in conjunction with this component.
+* Refer to [spacing](../guidelines/tokens/spacing) tokens to check possible string values for both width and height dimensions.
+* Refer to [colors](../guidelines/tokens/colors) tokens to check possible string values for both fill and stroke properties.
```
tcK1, line 19: Not sure if this is the right call, since we don't use tokens, we use the theme.

We have the `primary`, `secondary` and `tertiary` colours for exemple, which don't show up on the tokens page.

@ggdaltoso what do you think?

ggdaltoso: Tokens will always be references and what is missing is something to aim us on what we have available inside the theme that could help us. 
Since we don't have the theme documented yet, I'm sure this is enough for now.

 I also think the props table is pretty good:
propstable

We must revisit the entire documentation when we have the theme documentation. Until then, we could referrer to tokens as well


*_we don't have tertiary anymore_

ggdaltoso: We could link the [file that describes](https://github.com/Gympass/yoga/blob/alpha/packages/yoga/src/Theme/theme/theme.js#L8) the theme specification but I'm not sure it would help or not.

tcK1 (approved):

alan-oliv (approved):

ggdaltoso:
@all-contributors please add @coelhucas for code, ideas, doc

ggdaltoso:
@all-contributors please add @katharinedepaulo for ideas

PR #223, opened 2021-02-24
Feat/new tag

andrerocha22:
Updating `Tag` component according to Figma prototype [updates](https://www.figma.com/file/wn9grDi9mM29Wt7LmQjol7/Yoga-Design-System-2.0?node-id=4670%3A3942).


packages/yoga/src/Tag/Tag.theme.js
```diff
@@ -1,21 +1,24 @@
...
-    top: spacing.xxsmall,
+    top: spacing.xxxsmall,
     right: spacing.xxsmall,
-    bottom: spacing.xxsmall,
+    bottom: spacing.xxxsmall,
     left: spacing.xxsmall,
   },
   border: {
-    radius: radii.rounded,
+    width: borders.small,
```
ericcleao, line 20: @andrerocha22 Are we using this `width` attribute?

packages/yoga/src/Tag/native/Informative.jsx
```diff
@@ -0,0 +1,90 @@
...
+          text,
+          feedback: {
+            [variant]: { light = text.secondary },
+          },
+        },
+        components: { tag },
+      },
+    },
+  }) => `
+    background-color: ${hexToRgb(light)};
```
ericcleao, line 22: Do we need this `hexToRgb` call here? And if it is needed we don't need to use at other places like `border-color` and the colors from web Tags and default native Tag?

packages/yoga/src/Tag/native/Informative.jsx
```diff
@@ -0,0 +1,90 @@
...
+
+const Informative = styled(StyledTag)`
+  ${({
+    variant,
+    theme: {
+      yoga: {
+        colors: {
+          text,
+          feedback: {
+            [variant]: { light = text.secondary },
```
ericcleao, line 15: We have a pattern defined to this type of destruct? At some places we are using `[variant]: color = { light: text.secondary }` and others like this line, if we don't have a pattern defined we could just choose one and use it on all Tag components.

packages/yoga/src/Tag/native/Informative.jsx
```diff
@@ -0,0 +1,90 @@
...
+  </Informative>
+);
+
+TagInformative.propTypes = {
+  variant: oneOf(['', 'success', 'informative', 'attention']),
+  icon: oneOfType([bool, func]),
+  children: node.isRequired,
+};
+
+TagInformative.defaultProps = {
```
ericcleao, line 83: The default variant should not be the empty one?

packages/yoga/src/Tag/native/Tag.jsx
```diff
@@ -1,115 +1,77 @@
...
-    full,
+    variant,
     theme: {
       yoga: {
+        colors: {
+          text,
+          elements,
+          feedback: {
+            [variant]: color = { dark: text.secondary },
+            [variant]: borderColor = { dark: elements.selectionAndIcons },
```
ericcleao, line 19: We have a pattern defined to this type of destruct? At some places we are using [variant]: { light: = text.secondary } and others like this line, if we don't have a pattern defined we could just choose one and use it on all Tag components.

packages/yoga/src/Tag/native/Informative.jsx
```diff
@@ -0,0 +1,90 @@
+import React from 'react';
+import styled, { withTheme } from 'styled-components';
+import { func, oneOfType, oneOf, bool, node } from 'prop-types';
+import { hexToRgb } from '@gympass/yoga-common';
+import { StyledTag, StyledText } from './Tag';
```
ggdaltoso, line 5:
```suggestion

import { StyledTag, StyledText } from './Tag';
```

💅

packages/yoga/src/Tag/native/Informative.jsx
ggdaltoso, line 22:
```suggestion
    background-color: ${light};
```

Do we need this `hexToRgb`?

packages/yoga/src/Tag/web/Informative.jsx
```diff
@@ -0,0 +1,75 @@
...
+  justify-content: center;
+  align-items: center;
+
+  ${({
+    variant,
+    theme: {
+      yoga: {
+        colors: {
+          text,
+          feedback: { [variant]: color = { light: text.secondary } },
```
ericcleao, line 18: We have a pattern defined to this type of destruct? At some places we are using [variant]: { light = text.secondary } and others like this line, if we don't have a pattern defined we could just choose one and use it on all Tag components.

packages/yoga/src/Tag/native/Informative.jsx
```diff
@@ -0,0 +1,90 @@
...
+          marginRight={tag.icon.margin.right}
+        />
+      )}
+      <StyledTextInformative>{children}</StyledTextInformative>
+    </Wrapper>
+  </Informative>
+);
+
+TagInformative.propTypes = {
+  variant: oneOf(['', 'success', 'informative', 'attention']),
```
ggdaltoso, line 78:
```suggestion
  /** style the card following the theme (success, informative, attention) */
  variant: oneOf(['', 'success', 'informative', 'attention']),
```

📝

packages/yoga/src/Tag/web/Informative.jsx
```diff
@@ -0,0 +1,75 @@
...
+        },
+        components: { tag },
+      },
+    },
+  }) => `
+    background-color: ${hexToRgb(color.light)};
```
ericcleao, line 24: same comment about the `hexToRgb` method call from the native file

packages/yoga/src/Tag/native/Informative.jsx
```diff
@@ -0,0 +1,90 @@
...
+  }) => `
+    color: ${color};
+
+    font-size: ${tag.font.size}px;
+    line-height: ${tag.font.lineHeight}px;
+    font-weight: ${tag.font.weight};
+  `}
+`;
+
+const TagInformative = ({
```
ggdaltoso, line 51: We must have a description for this component. 
You can easily find it on its Figma file, in the `Definition` section.

packages/yoga/src/Tag/Tag.theme.js
andrerocha22: We're using here: `border: ${tag.border.width}px solid`

packages/yoga/src/Tag/web/Informative.jsx
ggdaltoso, line 24:
```suggestion
    background-color: ${color.light};
```

packages/yoga/src/Tag/web/Informative.jsx
```diff
@@ -0,0 +1,75 @@
...
+    }
+  `}
+`;
+
+const TagInformative = ({
+  children,
+  icon: Icon,
+  theme: {
+    yoga: {
+      colors: { primary, color = primary },
```
ggdaltoso, line 43:
```suggestion
      colors: { primary },
```

We don't need this `color` alias, right? 
Also, it should be better to check if the color is whether `primary` or `text.primary`

packages/yoga/src/Tag/web/Informative.jsx
```diff
@@ -0,0 +1,75 @@
...
+        fill={color}
+        marginRight={tag.icon.margin.right}
+      />
+    )}
+    {children}
+  </Informative>
+);
+
+TagInformative.propTypes = {
+  variant: oneOf(['', 'success', 'informative', 'attention']),
```
ggdaltoso, line 63:
```suggestion
  /** style the card following the theme (success, informative, attention) */
  variant: oneOf(['', 'success', 'informative', 'attention']),
```

📝

ggdaltoso (changes requested):
Awesome work on this component 👏

packages/yoga/src/Tag/native/Informative.jsx
andrerocha22: The _Tag.Informative_ does not have a default like _Tag_ , the default is the informative color. That is why we set a default variant.

packages/yoga/src/Tag/Tag.theme.js
ggdaltoso: since `border-style` is solid by default, we should change only what is needed to change. So, we should modify `border-width` instead.

packages/yoga/src/Tag/web/Tag.jsx
```diff
@@ -1,93 +1,58 @@
...
-    'deepPurple',
-    'stamina',
-    'dark',
-    'medium',
-    'deep',
-    'light',
-    'clear',
-    'white',
-  ]),
+  /** style the card following the theme (success, neutral, attention) */
```
ericcleao, line 50:
```suggestion
  /** style the tag following the theme (success, informative, attention) */
```

packages/yoga/src/Tag/native/Tag.jsx
```diff
@@ -1,115 +1,77 @@
...
-    'deepPurple',
-    'stamina',
-    'dark',
-    'medium',
-    'deep',
-    'light',
-    'clear',
-    'white',
-  ]),
+  /** style the card following the theme (success, informative, attention) */
```
ericcleao, line 69:
```suggestion
  /** style the tag following the theme (success, informative, attention) */
```

ericcleao (commented):
Great job, @andrerocha22!! 🚀

ericcleao:
@andrerocha22 and @ggdaltoso, now we are defining that the `TagInformative` variant could be empty but we are defining it's default as `informative`. We don't have a default variant on it different from the `Default Tag` that could have an empty `variant` passed and there are colors defined for it. At `TagInformative` we are letting the dev to use it as: `<TagInformative variant='' />` but it's not a valid variant, what do you think?

ggdaltoso:
> @andrerocha22 and @ggdaltoso, now we are defining that the `TagInformative` variant could be empty but we are defining it's default as `informative`. We don't have a default variant on it different from the `Default Tag` that could have an empty `variant` passed and there are colors defined for it. At `TagInformative` we are letting the dev to use it as: `<TagInformative variant='' />` but it's not a valid variant, what do you think?

It seems that `variant` prop in `Tag.Informative` is required. Right?

ericcleao (approved):

packages/yoga/src/Tag/web/Informative.jsx
```diff
@@ -0,0 +1,75 @@
...
+  },
+  ...props
+}) => (
+  <Informative {...props}>
+    {Icon && (
+      <Icon
+        width={tag.icon.size}
+        height={tag.icon.size}
+        fill={text.primary}
+        style={{ marginRight: tag.icon.margin.right }}
```
allyssonsantos, line 55:
```suggestion
```

You're already setting this `marginRight` on line `32` of this file, I think that you don't need this style here

allyssonsantos (changes requested):

packages/yoga/src/Tag/native/Tag.jsx
```diff
@@ -1,115 +1,74 @@
...
-          fill={color}
-          style={{ marginRight: tag.icon.margin.right + 1 }}
-        />
-      )}
-
-      <StyledText color={color}>{children}</StyledText>
-    </Wrapper>
+const Tag = ({ children, ...props }) => (
+  <StyledTag {...props}>
+    <StyledText {...props}>{children}</StyledText>
```
allyssonsantos, line 61:
```suggestion
const Tag = ({ children, variant, ...props }) => (
  <StyledTag variant={variant} {...props}>
    <StyledText variant={variant}>{children}</StyledText>
```

We will face some weird errors if you keep these two `{...props}` in `StyledTag` and `StyledText`.

Per exemple:
If someone wants to add a margin in this component, like this:
```jsx
<Tag style={{ marginRight: 28 }} />
```
It will be added on the `StyledTag` and in `StyledText`

allyssonsantos (changes requested):

packages/yoga/src/Tag/native/Tag.jsx
evilamaior: Done!

packages/yoga/src/Tag/web/Informative.jsx
evilamaior: Done!

allyssonsantos (approved):

ggdaltoso (approved):

ggdaltoso:
@all-contributors please add @andrerocha22 for code

PR #242, opened 2021-04-06
Update the PlanCard component to the new design

nypacheco:
- Update the PlanCard component to the new design. [Figma](https://www.figma.com/file/yuEFY338mVNHudQK3JX5Z3/%5BPixel%5D-Upgrade%2FDowngrade?node-id=6024%3A23394)
- Fix when the `buttonProps` (`PlanCard.ListItem`) does not exist and it renders a zero value.

| web | ios | android |
|:--- |:--- |:---     |
||||

### DIFF
--- a/packages/yoga/src/Card/native/PlanCard/Actions.jsx
+++ b/packages/yoga/src/Card/native/PlanCard/Actions.jsx
@@ -1,12 +1,40 @@
 import styled from 'styled-components';
 
 import Actions from '../Card/Actions';
+import Button from '../../../Button';
 import theme from '../../../Theme/helpers/themeReader';
 
-const Action = styled(Actions)`
-  margin-top: ${theme.components.card.plan.actions.margin.top}px;
+const {
+  components: {
+    card: {
+      plan: { actions },
+    },
+  },
+} = theme;
+
+const PlanCardButton = styled(Button).attrs({
+  small: true,
+  full: true,
+})``;
+
+const PlanCardButtonText = styled(Button.Text).attrs({
+  small: true,
+  full: true,
+})`
+  margin-top: ${actions.buttonText.margin.top}px;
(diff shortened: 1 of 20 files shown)


nypacheco:
Done 😄 

Talking to Geo, we update the Price margin `28px` to `32px` because the card that already exist have this same margin.

evilamaior (approved):

packages/yoga/src/Card/native/PlanCard/Actions.jsx
```diff
@@ -1,12 +1,40 @@
...
+
+const PlanCardButton = styled(Button).attrs({
+  small: true,
+  full: true,
+})``;
+
+const PlanCardButtonText = styled(Button.Text).attrs({
+  small: true,
+  full: true,
+})`
```
coelhucas, line 23: Do we have any criteria on when do we use `attrs` and when do we just simply send these props directly in `<PlanCardButton small full />`?

(It's not change request, really just a doubt)

coelhucas (approved):
You're an angel, thanks for that! I left a comment but in overall LGTM.

packages/yoga/src/Card/native/PlanCard/Actions.jsx
nypacheco: I put `attrs` in this case because the `PlanCard.Button` and `PlanCard.ButtonText` always be small and full, then the user of the component will not need to add this `attrs` in `<PlanCard.Button small full>` every time she/he uses it. 

Do you think it is better to leave it to the user to add this manually?

ericcleao (approved):

allyssonsantos (approved):

allyssonsantos:
@all-contributors please add @nypacheco for code, doc

PR #250, opened 2021-05-05
Fix AutoComplete onClean/onSelect event

allyssonsantos:
This PR fixes the issue that "onSelect" event was being called when the user clicks on the "Clear (X)" button.

|Before|After|
|------|-----|
|autocomplete-before|autocomplete-after|

### DIFF
--- a/packages/yoga/src/AutoComplete/web/AutoComplete.jsx
+++ b/packages/yoga/src/AutoComplete/web/AutoComplete.jsx
@@ -159,14 +159,19 @@ const AutoComplete = ({
     <Downshift
       selectedItem={userValue}
       onStateChange={changes => {
-        if (Object.prototype.hasOwnProperty.call(changes, 'selectedItem')) {
-          setUserValue(changes.selectedItem);
-          onSelect(changes.selectedItem);
+        const { selectedItem, inputValue } = changes;
+        if (
+          Object.prototype.hasOwnProperty.call(changes, 'selectedItem') &&
+          selectedItem
+        ) {
+          setUserValue(selectedItem);
+          onSelect(selectedItem);
+          onChange(selectedItem);
         } else if (
           Object.prototype.hasOwnProperty.call(changes, 'inputValue')
         ) {
-          setUserValue(changes.inputValue);
-          onChange(changes.inputValue);
+          setUserValue(inputValue);
+          onChange(inputValue);
         }
       }}
     >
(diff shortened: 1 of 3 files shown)


coelhucas (approved):
Well documented :)

packages/yoga/src/AutoComplete/web/AutoComplete.jsx
```diff
@@ -159,14 +159,19 @@ const AutoComplete = ({
     <Downshift
       selectedItem={userValue}
       onStateChange={changes => {
-        if (Object.prototype.hasOwnProperty.call(changes, 'selectedItem')) {
-          setUserValue(changes.selectedItem);
-          onSelect(changes.selectedItem);
+        const { selectedItem, inputValue } = changes;
+        if (
+          Object.prototype.hasOwnProperty.call(changes, 'selectedItem') &&
+          selectedItem
```
tcK1, line 165: Is there a difference between `Object.prototype.hasOwnProperty.call(changes, 'selectedItem')` and `Boolean(changes.selectedItem)` in this case? Same applies to the `inputValue` validation.

What would happen if we just used then directly?

```js
if (selectedItem) {
  /* ... */
} else if (inputValue) {
  /* ... */
}
```

packages/yoga/src/AutoComplete/web/AutoComplete.test.jsx
```diff
@@ -116,6 +124,33 @@ describe('<AutoComplete />', () => {
       fireEvent.click(getByRole('button'));
 
       expect(onCleanMock).toHaveBeenCalledWith('');
+      expect(onSelectMock).not.toHaveBeenCalled();
+      expect(onChangeMock).toHaveBeenCalled();
```
tcK1, line 128: Should we use `toHaveBeenCalledWith` on these specs to assert if we are passing the correct values to the callback functions?

allyssonsantos: yes! well noticed, thanks!

packages/yoga/src/AutoComplete/web/AutoComplete.jsx
allyssonsantos: When a user types or selects some suggestion on the field, `changes` object will have "selectedItem" and/or "inputValue" keys inside it.

Talking about "selectedItem" it doesn't make any difference using hasOwnProperty or not.

When we talk about `inputValue` we must call `onChange` even if the "inputValue" is an empty string (like if the user clicks on the "clean" button or deletes the whole string). So, we need to make sure that "inputValue" key exists (even with a falsy value).

tcK1 (approved):
🚀

packages/yoga/src/AutoComplete/web/AutoComplete.jsx
tcK1: That's perfect.

PR #256, opened 2021-05-27
Remove topic 'media.min' at media

Dario-Felipe:
The 'media.min' is present in the documentation, but it is redundant since the topic above has the same function. In addition, using this media in the code will give error for not recognizing the 'min'.

image

The correct way to use it would be, per example:

```
${media.md``}
```
or
```
${media.not['md']``}
```


allyssonsantos (approved):

ggdaltoso (approved):

caioalexandrebr (approved):
🚀

ggdaltoso (approved):

allyssonsantos:
@all-contributors please add @Dario-Felipe for doc

PR #261, opened 2021-06-02
New Component: Chips

ggdaltoso:
## Chips!

```jsx
import { Chips } from '@gympass/yoga';

<>
  <Chips onPress={onToggleClass} selected={classes}>
    Classes
  </Chips>
  <Chips disabled>Gyms and studios</Chips>
  <Chips selected counter={1350}>
    Personal trainers
  </Chips>
  <Chips icons={[MapPin]}>Activities</Chips>
  <Chips icons={[Filter]} />
  <Chips selected icons={[MapPin, ChevronDown]}>
    Availability
  </Chips>
</>
```

image


allyssonsantos (approved):

matheusmartins10 (approved):

coelhucas (approved):
I'd only suggest to add an example sending just left icon, just in case so it's explicit that people can do that without seeing the source code
```jsx
<Chips icon={[, Filter]}>Flip Flop</Chips>
```

packages/doc/content/components/components/chips/chips-web.mdx
```diff
@@ -0,0 +1,29 @@
...
+
+```javascript state
+render(() => {
+  const [first, toggleFirst] = useState(false);
+  const [second, toggleSecond] = useState(false);
+  const onToggleFirst = () => toggleFirst(!first);
+  const onToggleSecond = () => toggleSecond(!second);
+
+  return (
+    <>
```
caioalexandrebr, line 15: Very good contribution! :rocket:

This line will break in codesandbox, how about adding a `React.Fragment` or a `div` like you did in **Button.Default**?

https://gympass.github.io/yoga/components/button/default#default

Dario-Felipe: Hey everyone, studying this problem and from other codes written in CodeSandBox, this problem can be solved throughout Yoga by placing "react-scripts" as a dependency, as in the image below:

image

And it will look like that in CodeSandBox: 

image

I would like to hear from the team what they think about the implementation of this new dependency.
If it is consistent, I will open a PR with the correction. 🚀🚀

caioalexandrebr (approved):

packages/doc/content/components/components/chips/chips-web.mdx
Dario-Felipe: Link to [pull request](https://github.com/Gympass/yoga/pull/263) for possible solution.

andrerocha22 (approved):

allyssonsantos:
> I'd only suggest to add an example sending just left icon, just in case so it's explicit that people can do that without seeing the source code
> 
> ```js
> <Chips icon={[, Filter]}>Flip Flop</Chips>
> ```

Done @coelhucas!

Added an example like this:
image

PR #262, opened 2021-06-02
Fix: import of styled components

joaovicdsantos:
# Import styled components

image

image

### DIFF
--- a/packages/doc/src/components/CodeBlock/CodeBlock.jsx
+++ b/packages/doc/src/components/CodeBlock/CodeBlock.jsx
@@ -12,9 +12,11 @@ import {
 
 const buildImportString = (code, modules) => {
   const findComponents = /(?:<|{)(\w*)(?=\s*?\/?>*)/gm;
+  const findStyledComponents = /styled\(\w*/gm;
   const sortModules = /(@gympass\/yoga*)/gm;
   const imports = [];
   const foundComponents = code.match(findComponents) || [];
+  const foundStyledComponents = code.match(findStyledComponents) || [];
 
   modules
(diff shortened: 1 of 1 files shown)


ggdaltoso:
Hi, @joaovicdsantos ! Thanks for contributing!

Despite your contribution seems to be a fix, if we merge your PR into master, we will trigger a new version of Yoga which isn't your purpose, right?. 

I would ask you to change your commit message to be `docs` or `chore` instead of `fix`. Just to prevent Yoga bumps.

| Before | After |
| ------- | ----- |
| ```fix(codeblock): fix import of styled elements``` | ```docs(codeblock): fix import of styled elements```  <br /> or <br /> ```chore(codeblock): fix import of styled elements``` |

joaovicdsantos:
Hi, @ggdaltoso !

Link of the new PR with the requested changes: https://github.com/Gympass/yoga/pull/265

coelhucas:
Hey, @joaovicdsantos. Letting you know, just in case: you could had done these changes without closing your original PR. You could have replaced the commit's name by using [`commit --amend`](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---amend). [Here's](https://www.atlassian.com/git/tutorials/rewriting-history) a small tutorial.

If you had multiple commits to be renamed, you could also use [`rebase -i`](https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt--i).  There's also some reference [here](https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase) (look for `interactive` mode, in this case).

I hope to see more of your contributions around! :)

PR #263, opened 2021-06-03
Add 'react-scripts' as dependency

Dario-Felipe:
It was observed that when a code is generated in CodeSandBox, where there is a `<></> (fragment)` it will give a certain error. Here is the error example:

image

One of the solutions would be to put in each component that uses it `<React.Fragment> or <div>`. As in the example below:

image

However, it can be solved by adding the dependency 'react-scripts', and that way the `<></> (fragment)` will be recognized in all code generated in CodeSandBox from the Yoga documentation. Example:

image

### DIFF
--- a/packages/doc/src/components/CodeBlock/ReactLive/sandboxConfig.js
+++ b/packages/doc/src/components/CodeBlock/ReactLive/sandboxConfig.js
@@ -11,6 +11,7 @@ const getPackage = ([imports, component]) =>
           dependencies: {
             react: 'latest',
             'react-dom': 'latest',
+            'react-scripts': 'latest',
             '@gympass/yoga': 'latest',
             'styled-components': 'latest',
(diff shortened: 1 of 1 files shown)


ggdaltoso:
Hi, @Dario-Felipe ! Thanks for contributing again!

Despite your contribution seems to be a fix, if we merge your PR into master, we will trigger a new version of Yoga which isn't your purpose, right?. 

I would ask you to change your commit message to be `docs` instead of `fix`. Just to prevent Yoga bumps.

| Before | After |
| ------- | ----- |
| ```fix(codeblock): add 'react-scripts' as dependency``` | ```docs(codeblock): add 'react-scripts' as dependency``` |

ggdaltoso (approved):

caioalexandrebr (approved):

PR #264, opened 2021-06-04
Yoga system

ggdaltoso:
image

The new Yoga System pretends to make the usage of our theme easy as it should be.

Each package module adds a way to use our decisions by just adding the respective module you want, like:

```jsx
import { spacing } from '@gympass/yoga-system';

const Square = styled.div`
  border: 1px solid;

  ${spacing}
`;

<Square ph="medium" paddingVertical="small" mb="medium">
  Everyone can be happy and healthy
</Square>
```

In the example above, we are adding the spacing module into `Square` component which also adds all space props into it :tada:

The modules inside this package are:

- `spacing` (having all bellow)
- `margin`
- `marginTop`
- `marginRight`
- `marginBottom`
- `marginLeft`
- `marginHorizontal`
- `marginVertical`
- `padding`
- `paddingTop`
- `paddingRight`
- `paddingBottom`
- `paddingLeft`
- `paddingHorizontal`
- `paddingVertical`

<br />

- `border` (having all bellow)
- `borderTop`
- `borderRight`
- `borderBottom`
- `borderLeft`
- `borderColor`
- `borderTopColor`
- `borderRightColor`
- `borderBottomColor`
- `borderLeftColor`
- `borderWidth`
- `borderTopWidth`
- `borderRightWidth`
- `borderBottomWidth`
- `borderLeftWidth`
- `borderRadius`
- `borderTopLeftRadius`
- `borderTopRightRadius`
- `borderBottomLeftRadius`
- `borderBottomRightRadius`

<br />

- `backgroundColor`

<br />

- `elevation`

<br />

- `typography` (having all bellow)
- `color`
- `fontSize`
- `fontWeight`
- `lineHeight`

<br />

In addition to that, we have the `system`, including all props to your component.


packages/system/README.md
```diff
@@ -0,0 +1,11 @@
+# `@gympass/yoga-system`
+
+> TODO: description
+
+## Usage
+
+```
+const system = require('@gympass/yoga-system');
+
+// TODO: DEMONSTRATE API
```
Wendlereis, line 10: Is it missing?

packages/system/src/unit.js
```diff
@@ -0,0 +1,6 @@
+const toUnit = (unit = 'px') => value =>
+  Number(value) && value !== 0 ? `${value}${unit}` : value;
+
+const toPx = toUnit('px');
```
Wendlereis, line 4: ❤️

packages/system/README.md
ggdaltoso: Yes! Thanks!

Wendlereis (approved):

packages/system/src/theme.js
```diff
@@ -0,0 +1,86 @@
...
+    },
+  } = props;
+
+  return matchedSpec;
+};
+
+const getSpacing = props => getFromTheme(props)('spacing');
+const getBorder = props => getFromTheme(props)('borders');
+const getBorderRadius = props => getFromTheme(props)('radii');
+const getColor = props => getFromTheme(props)('colors');
```
tcK1, line 27: I may be missing something, but shouldn't this be `text.colors` (or something like that haha)? 🤔

ggdaltoso: by getting from `colors.text` instead of `colors`, we won't be able to use all colors such as Vibin, Hope, Energy, and so on.

Getting from colors, we can get all of them plus all text colors, simply by doing:

```jsx
import { color } from '@gympass/yoga-system';

const Text = styled.p(color);

{/* works */}
<Text color="hope" /> 

{/* works too */}
<Text color="text.primary" /> 
```

So, with this in mind, I would rather get from `colors` to only `colors.text`.
What do you think?

packages/system/src/theme.js
```diff
@@ -0,0 +1,86 @@
...
+    const v = prop
+      .map(p => generator({ props, prop: p, cssProperty, getter, transform }))
+      .flat();
+
+    return v;
+  }
+
+  const p = props[prop];
+
+  const value = resolve(themeProp, p) || p;
```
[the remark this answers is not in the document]
ggdaltoso: Is it better now?

packages/system/src/theme.js
tcK1: This decision has to be made by the design team, doesn't it? Should we be able to have texts on all these colors or should we limit what we can do?

If that's what we want then I don't mind. 😁

ggdaltoso: We have the ok from the PD team. We're all done, so. Ok?

tcK1 (approved):

coelhucas (approved):
Excited to see it working

New Project (2)

PR #265, opened 2021-06-09
Doc: import of styled components

joaovicdsantos:
# Import styled components

image

image


joaovicdsantos:
**Hey what are you doing here?** 
Pretend nothing happened and follow the link below to access the correct PR: 😆 

https://github.com/Gympass/yoga/pull/266

PR #266, opened 2021-06-09
Doc: fix import of styled elements

joaovicdsantos:
# Import styled components

image

image


allyssonsantos (approved):

caioalexandrebr (approved):

ggdaltoso (approved):

ggdaltoso:
@all-contributors please add @joaovicdsantos for doc

PR #275, opened 2021-06-22
Fix Android font weight issue

thiagogympass:
There's a font weight issue on [Android](https://github.com/facebook/react-native/issues/25696)
`font-weight` style props does not work on Android.

This approach will work with `Text` and its variations.

There are many components on Yoga that are using `font-weight` and will need fix.

| Before | After |
| ------ | ----- |
|Captura de Tela 2021-06-21 às 15 06 51|Captura de Tela 2021-06-22 às 10 55 45|


coelhucas (approved):

coelhucas:
Thanks for looking this!

katharinedepaulo (approved):

ggdaltoso (approved):

ggdaltoso:
@all-contributors please remove @thiagogympass and add @tjamancio for design and bug

PR #284, opened 2021-06-30
Upgrade system

ggdaltoso:
image

Keeping the work in `@gympass/system`, we're adding the follow new modules:

- `flexes` (having all bellow)
  - `flexBasis`
  - `flexFlow`
  - `flexGrow`
  - `flexShrink`
  - `flexWrap`
  - `flexDirection`
  - `alignItems`
  - `alignContent`
  - `alignSelf`
  - `justifyContent`
  - `justifySelf`
  - `order`

<br />

- `positions` (having all bellow)
   - `position`
   - `top`
   - `right`
   - `bottom`
   - `left`

<br />

- `display`

<br />

Also, we're adding to `spacing`:

 - `width`
 - `height`

Those updates will be reflected into our Box component as it makes use of the `system` module


renannobile (approved):

mmartinsoliveira (approved):

packages/system/src/flex.js
```diff
@@ -0,0 +1,126 @@
...
+    prop: ['flexWrap'],
+    cssProperty: 'flexWrap',
+  });
+
+const order = props =>
+  generator({
+    props,
+    prop: ['order'],
+    cssProperty: 'order',
+    getter: () => ({}),
```
alan-oliv, line 92:
```suggestion
```
this getter isn't default?

alan-oliv (approved):

PR #287, opened 2021-07-14
New component avatar

LeticiaSoares:
I created the component **Avatar** in yoga-labnative and I added a icon in yoga-icons.For a while we have only the component in native.

https://www.figma.com/file/wn9grDi9mM29Wt7LmQjol7/Yoga-Design-System-7.1?node-id=4931%3A664

image


packages/yoga/src/Avatar/native/Avatar.jsx
```diff
@@ -0,0 +1,60 @@
...
+  src: undefined,
+  placeholder: undefined,
+  fill: undefined,
+  width: undefined,
+  height: undefined,
+  borderRadius: undefined,
+  type: undefined,
+};
+
+Avatar.displayName = 'Avatar';
```
leonardokl, line 58: Why is this value being assigned? This is the same value that will be inferred by react. I saw that this is done in other places but I didn't understand the motivation.

packages/yoga/src/Avatar/native/Circle.jsx
```diff
@@ -0,0 +1,39 @@
...
+
+Circle.defaultProps = {
+  src: undefined,
+  placeholder: undefined,
+  fill: undefined,
+  width: undefined,
+  height: undefined,
+};
+
+Circle.displayName = 'Avatar.Circle';
```
leonardokl, line 37: The component could be named `AvatarCicle`, so the inferred displayName would be sufficient to describe the component.

packages/yoga/src/Avatar/native/Avatar.jsx
```diff
@@ -0,0 +1,60 @@
...
+const Avatar = ({
+  src,
+  fill,
+  placeholder,
+  width = 48,
+  height = 48,
+  borderRadius = 'small',
+  type = 'default',
+}) => {
+  const icon = placeholder || BuildingFilled;
```
leonardokl, line 17: Could `BuildingFilled` be the default value for `placeholder`?

```js
const Avatar = ({
  src,
  fill,
  placeholder = BuildingFilled,
  width = 48,
  height = 48,
  borderRadius = 'small',
  type = 'default',
}) => {
```

packages/yoga/src/Avatar/native/Avatar.jsx
```diff
@@ -0,0 +1,60 @@
...
+
+Avatar.defaultProps = {
+  src: undefined,
+  placeholder: undefined,
+  fill: undefined,
+  width: undefined,
+  height: undefined,
+  borderRadius: undefined,
+  type: undefined,
+};
```
leonardokl, line 56: Do we need to set the defaultProps? The `Avatar` parameters already have default values.
I think we should use one or the other. I prefer the function parameters as they end up helping in Intelisense.

```js
const Avatar = ({
  src,
  fill,
  placeholder,
  width = 48,
  height = 48,
  borderRadius = 'small',
  type = 'default',
})
````

OR

```js
Avatar.defaultProps = {
  src: undefined,
  placeholder: undefined,
  fill: undefined,
  width: 48,
  height: 48,
  borderRadius: 'small',
  type: 'default',
};
```

packages/yoga/src/Avatar/native/Placeholder.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import { func, string } from 'prop-types';
+import Icon from '../../Icon';
+
+const Placeholder = ({ fill, icon }) => {
+  return <Icon as={icon} width="50%" height="50%" fill={fill || 'white'} />;
```
leonardokl, line 6: Could `white` be used as the default value of `fill`?

```js
const Placeholder = ({ fill = 'white', icon }) => {
```

OR

```js
Placeholder.defaultProps = {
  fill: 'white',
};
```

packages/yoga/src/Avatar/native/Circle.jsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import { UserFilled } from '@gympass/yoga-icons';
+import { func, string, number } from 'prop-types';
+import Avatar from './Avatar';
+
+const Circle = ({ src, fill, placeholder, width, height }) => {
+  const icon = placeholder || UserFilled;
```
leonardokl, line 7: Could `UserFilled ` be used as the default value of `placeholder`?

```js
const Circle = ({ src, fill, placeholder = UserFilled, width, height }) => {
```

OR

```js
Circle.defaultProps = {
  placeholder: UserFilled,
};
```

packages/yoga/src/Avatar/native/Avatar.jsx
LeticiaSoares: I also don't know the motivation. Can you tell us? @ggdaltoso

packages/yoga/src/Avatar/native/Avatar.jsx
```diff
@@ -0,0 +1,60 @@
...
+  return (
+    <Box
+      bgColor="elements.selectionAndIcons"
+      display="flex"
+      width={width}
+      height={height}
+      alignItems="center"
+      justifyContent="center"
+      elevation="small"
+      borderRadius={borderRadius}
```
leonardokl, line 27: We could spread the other properties here, this way it would make it easier to customize the component and add other properties. The same goes for `Circle`.

```js
borderRadius={borderRadius}
{...otherProps}
```

packages/yoga/src/Avatar/native/Avatar.jsx
LeticiaSoares: @leonardokl  I prefer too using in parameters, but the lint of project required have defaultProps.

packages/yoga/src/Avatar/native/Avatar.jsx
LeticiaSoares: @leonardokl Done!

packages/yoga/src/Avatar/native/Circle.jsx
LeticiaSoares: @leonardokl Done!

packages/yoga/src/Avatar/native/Placeholder.jsx
LeticiaSoares: @leonardokl Done!

packages/yoga/src/Avatar/native/Avatar.jsx
LeticiaSoares: @leonardokl Done! I added in default props only this values

packages/yoga/src/Avatar/native/Avatar.jsx
LeticiaSoares: @leonardokl Done!

packages/yoga/src/Avatar/native/Circle.jsx
LeticiaSoares: I changed the name of component to AvatarCircle

rafaelcoleta (approved):

packages/yoga/src/Avatar/native/Avatar.jsx
LeticiaSoares: @leonardokl the @ggdaltoso is on vacation.Is it ok we ask about this when he returnerd?

leonardokl (approved):

alycecristines (approved):

packages/yoga/src/Avatar/Avatar.theme.js
```diff
@@ -0,0 +1,3 @@
+const Avatar = () => ({});
```
alan-oliv, line 1: Hi Le,
If you don't see the need to themify the avatar, you don't need this file
You can look at the 'chip' component, it doesn't have a theme file because it inherits all from base

PR #299, opened 2021-08-05
Bump React Native to 0.63.4

ggdaltoso:


packages/labnative/android/app/src/debug/java/com/labnative/ReactNativeFlipper.java
```diff
@@ -0,0 +1,72 @@
...
+import com.facebook.flipper.plugins.network.FlipperOkhttpInterceptor;
+import com.facebook.flipper.plugins.network.NetworkFlipperPlugin;
+import com.facebook.flipper.plugins.react.ReactFlipperPlugin;
+import com.facebook.flipper.plugins.sharedpreferences.SharedPreferencesFlipperPlugin;
+import com.facebook.react.ReactInstanceManager;
+import com.facebook.react.bridge.ReactContext;
+import com.facebook.react.modules.network.NetworkingModule;
+import okhttp3.OkHttpClient;
+
+public class ReactNativeFlipper {
```
tcK1, line 27: ❤️

packages/yoga/src/Dropdown/native/Options.android.jsx
```diff
@@ -46,31 +46,30 @@ const OptionText = styled.Text`
 `;
 
 const Options = ({ options, selectedOption, onSelect }) => (
-  <ScrollView contentContainerStyle={{ width: '100%', height: 60 * 3.6 }}>
```
tcK1, line 49: We don't need this `ScrollView` anymore? Has the behaviour changed or we didn't need it at all?

ggdaltoso: We don't because its children is a FlatList Component.

The FlatList renders only the visible elements (kinda VirtualList thing) and the ScrollView will always render everything. Putting it in that order, FlatList as a child of ScrollView, the React Native will fire a warning saying this:

```
VirtualizedLists should never be nested inside plain ScrollViews with the same orientation
 - use another VirtualizedList-backed container instead
```

So, I removed that ScrollView, and everything keeps working as it should. ✨

tcK1 (approved):

katharinedepaulo (approved):

PR #302, opened 2021-08-12
New component Avatar for web

mmartinsoliveira:
# ✨ Add component Avatar web

image

### Checklist

- [x] Unit tests

Co-authors: @ggdaltoso @caioalexandrebr @LeticiaSoares


coelhucas:
Do you folks pretend adding this to native as well? It has lot of usage there in our app too.

LeticiaSoares:
> Do you folks pretend adding this to native as well? It has lot of usage there in our app too.

I already have the version for native.

packages/yoga/src/Avatar/web/Avatar.jsx
```diff
@@ -0,0 +1,68 @@
...
+    borderRadius="small"
+    elevation="small"
+    overflow="hidden"
+    {...otherProps}
+  >
+    {src ? (
+      <img src={src} alt={alt} />
+    ) : (
+      <Icon
+        as={placeholder}
```
tcK1, line 25: What do we think about calling this `placeholder`? When I read it I thought about the Input placeholder attribute.

What do you think about naming it `icon`?

**Before:**
```jsx
import { YouTube } from '@gympass/yoga-icons';

<Avatar placeholder={YouTube} />
```

**After:**
```jsx
import { YouTube } from '@gympass/yoga-icons';

<Avatar icon={YouTube} />
```

packages/yoga/src/Avatar/web/Avatar.jsx
```diff
@@ -0,0 +1,68 @@
...
+        stroke={stroke}
+      />
+    )}
+  </Box>
+);
+
+Avatar.propTypes = {
+  src: string,
+  /**  Required if src is true */
+  alt: (props, propName, componentName) => {
```
tcK1, line 38: Loved this concept! Amazing idea.

Maybe we should apply the same logic to the combination of `src` and `placeholder` props, since we can only use one of them at a time.

packages/yoga/src/Avatar/web/AvatarCircle.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { UserFilled } from '@gympass/yoga-icons';
+import Avatar from './Avatar';
+
+const AvatarCircle = props => (
+  <Avatar {...props} borderRadius="circle" placeholder={UserFilled} />
```
tcK1, line 6: This way we can't use anything but the `UserFilled` icon with the `Avatar.Circle` component. Do we even need this "default" behavior?

Even the scenario on our spec also uses the same icon and that spec only passes because it is a snapshot matching.

mmartinsoliveira: You're absolutely right! Thanks

packages/yoga/src/Avatar/web/Avatar.jsx
```diff
@@ -0,0 +1,74 @@
...
+import Icon from '../../Icon';
+
+const Avatar = ({ src, alt, fill, stroke, icon, ...otherProps }) => (
+  <Box
+    bgColor="elements.selectionAndIcons"
+    display="flex"
+    alignItems="center"
+    justifyContent="center"
+    width={48}
+    height={48}
```
tcK1, line 15: Shouldn't these be theme values?

packages/yoga/src/Avatar/web/AvatarCircle.jsx
ggdaltoso: Our design guidelines show us that when we are using `Avatar.Circle` the placeholder must be the `UserFilled`.With this in mind, we are defining the `placeholder` and `borderRadius` to reflect those decisions. 
Any other Avatar you may need, you can achieve it by just using Avatar and them providing some props.

caioalexandrebr (approved):

coelhucas (approved):
Don't we need variations for the size? I feel like we have (at least in app) different sizes for this kind of component. 

cc/ @tcK1 @naabraz @nypacheco

Wendlereis (approved):
Flawless 🙌

tcK1 (approved):

alycecristines (approved):

ggdaltoso:
@all-contributors please add @mmartinsoliveira and @caioalexandrebr for code and ideas

ggdaltoso:
@all-contributors please add @caioalexandrebr for code and ideas

ggdaltoso:
@all-contributors please add @marcosricardo for doc

PR #307, opened 2021-08-13
Add size prop to Icon component

Falkaniere:
# Add size prop to Icon component

Before: 
```jsx
<Icon as={AlertCircle} width="medium" height="medium" fill="energy" />
```

After: 
```jsx
<Icon as={AlertCircle} size="medium" fill="energy" />
```


packages/yoga/src/Icon/Icon.jsx
```diff
@@ -1,11 +1,20 @@
...
+} from 'prop-types';
 import { withTheme } from 'styled-components';
 
 const Icon = ({
   as: Component,
-  width,
-  height,
+  size,
+  width = size,
+  height = size,
```
tcK1, line 17: Clever! 🚀

packages/yoga/src/Icon/Icon.jsx
```diff
@@ -49,13 +58,38 @@ Icon.propTypes = {
...
+      checkPropTypes(
+        {
+          [propName]: oneOfType([oneOf(commonSizes), string, number]),
+        },
+        { size },
+        'prop',
+        componentName,
+      );
+
+      if (width || height) {
```
tcK1, line 76: Maybe this condition should come first, instead of nested conditionals:

```js
if (width || height) throw new Error(/* ... */);

checkPropTypes(/* ... */)
```

ggdaltoso (approved):
Small change for a big improvement!

Thanks, @Falkaniere

packages/yoga/src/Icon/Icon.jsx
```diff
@@ -49,13 +58,38 @@ Icon.propTypes = {
   /** Vertical size of the SVG. Use it as one of
    * theme.spacing tokens (xxsmall, small, medium...) */
   height: oneOf([...commonSizes, instanceOf(Number)]),
+  /** Size for vertical and horizontal of the SVG.
+   * Use it as one of theme.spacing tokens (xxsmall, small, medium...) */
+  size: (props, propName, componentName) => {
+    const { size, width, height } = props;
+
+    if (width || height) {
+      return new Error(
```
tcK1, line 67: Sorry, I don't really know how this prop type validation really works, but we usually `throw` errors, should we change here? Do we really need to return?

packages/yoga/src/Icon/Icon.jsx
```diff
@@ -49,13 +58,38 @@ Icon.propTypes = {
...
+        `you must use only ${propName}, alone, or width and/or height in ${componentName}`,
+      );
+    }
+
+    if (size) {
```
tcK1, line 72: Do we need this conditional? 🤔

packages/yoga/src/Icon/Icon.jsx
ggdaltoso: If someone uses `size` they don't need to use `width` or `height`. I think we just need to add `size` as a conditional too.

Talking about returning new Error instead of throwing it, we're just following the guidelines from [`prop-types` API](https://reactjs.org/docs/typechecking-with-proptypes.html):

```jsx
  // You can also specify a custom validator. It should return an Error
  // object if the validation fails. Don't `console.warn` or throw, as this
  // won't work inside `oneOfType`.
  customProp: function(props, propName, componentName) {
    if (!/matchme/.test(props[propName])) {
      return new Error(
        'Invalid prop `' + propName + '` supplied to' +
        ' `' + componentName + '`. Validation failed.'
      );
    }
  },
```

tcK1 (approved):

packages/yoga/src/Icon/Icon.jsx
willianpossa: In that case, if the developer don't pass size and width && height are empty, why we doesn't set a defaultValue following size guide to device? (Web, Native and etc).

willianpossa (approved):

packages/yoga/src/Icon/Icon.jsx
ggdaltoso: If you don't provide any size prop (`width`, `height`, or `size`) the Icon will render as 12x12 px, as it is the default value for all Icon on [@gympass/yoga-icons](https://github.com/Gympass/yoga/blob/master/packages/icons/src/svg/accessibility.svg?short_path=7d5248a#L1). But, sure, we can provide a default size if it is a thing.

I also think we should discuss this subject despite this pull request and add this potential default behavior later. Is it ok for you, @willianpossa ?

packages/yoga/src/Icon/Icon.jsx
ggdaltoso: We don't! Fixed!

packages/yoga/src/Icon/Icon.jsx
willianpossa: Of course @ggdaltoso!

I believe this behavior is a feature that we can implement later (Get size according device used)

willianpossa (approved):

ggdaltoso:
@all-contributors please add @Falkaniere for code and ideas

PR #318, opened 2021-08-23
Elevation. Revisited.

ggdaltoso:
# Breaking change
It isn't 😄 

Our elevation system has never worked as expected on Android devices. We always needed to add `elevation` and `box-shadow` to the component and try to figure out how shadow it would generate on iOS and Android as well.

This PR changes our tokens, adding the specifics for each platform.

### Web

| Before | After |
| ------ | ----- |
| image | image |

### Android

image

### iOS

image

It is important to say that those changes on our system will be reflected in our Box component, making it possible to use the same code to add our shadows to a component:

| Code | Web | Android | iOS |
| ----- | ----- | -------- | --- |
| image | image | image | image |


packages/yoga/src/Card/native/PlanCard/__snapshots__/PlanCard.test.jsx.snap
```diff
@@ -11,51 +11,57 @@ exports[`<PlanCard /> Snapshots should match snapshot with default PlanCard 1`]
...
   >
     <View
+      bRadius={8}
+      bgColor="#FFFFFF"
+      elevation="0 2px 4px      rgba(152, 152, 166, 0.2)"
+      overflow="hidden"
+      pb={16}
+      pl={16}
+      pr={16}
+      pt={16}
```
tcK1, line 37: 🕵🏼

tcK1 (changes requested):
Instead of using the `.attrs()` method, what do you think of this approach:

```jsx
// Using the card element as an example

<Box elevation="components.card.elevation" {...props} />
```

Would it work? This can be used a lot on this PR. 🤔 

-------

It seems like we are passing props directly to the `View` component in some cases. I've marked a single occurrence as an example, but there are many on this PR. We should find where we are doing it and fix it.

andreloureiro (approved):

coelhucas (approved):

ggdaltoso:
Good point, @tcK1 

For the first, your suggestion wouldn't work because our system matches each prop to its respective prop on `theme`. For example:

This works because we have a color, on our theme, matching `feedback.success.dark` 

<table>
<tr>
<td>

  ```jsx
<>
    <Box color="feedback.success.light" />  
    <Box bgColor="feedback.success.dark" />  
</>
```
</td>
<td>

  ```js
theme: {
    yoga: {
      colors: {
        success: {
          light: "#C1EEDB"
          dark: "#1D856C"
        }
      },
    }
}
```
</td>
</tr>
</table>

So, with this in mind, we don't match the entire theme to the prop itself. Just the respective ones. And this is why your code wouldn't work:




<table>
<tr>
<td>

```jsx
// Using the card element as an example
<Box elevation="components.card.elevation" {...props} />
```
</td>
<td>

  ```js
theme: {
    yoga: {
      elevations: {
        0: "none"
        1: "0 2px 6px rgba(152, 152, 166, 0.25)",
        2: "0 4px 12px rgba(152, 152, 166, 0.25)",
        3: "0 8px 20px rgba(152, 152, 166, 0.25)",
        zero: "none",
        small: "0 2px 6px rgba(152, 152, 166, 0.25)",
        medium: "0 4px 12px rgba(152, 152, 166, 0.25)",
        large: "0 8px 20px rgba(152, 152, 166, 0.25)",
      }
    }
}
```
</td>
</tr>
</table>

You are trying to match the elevation prop to the components but our system will try to resolve it as an elevation and it would not find the respective.

ggdaltoso:
> Secondly, it seems like we are passing props directly to the `View` component in some cases. I've marked a single occurrence as an example, but there are many on this PR. We should find where we are doing it and fix it.

For this second point, I think it is a [styled-components issue](https://github.com/styled-components/styled-components/issues/135) that has been solved (?) in v5 with [shouldforwardprop](https://styled-components.com/docs/api#shouldforwardprop) and it could be revisited when we bump SC version. Is it good for you, @tcK1 ?

tcK1:
> Is it good for you, @tcK1 ?

Okay!

tcK1 (approved):

alycecristines (approved):

PR #321, opened 2021-08-24
Allow UserFilledStar to have a dynamic color

leonardokl:
It was only possible to use the default icon color.

Screen Shot 2021-08-24 at 17 05 28
Screen Shot 2021-08-24 at 17 06 06


rafaelcoleta (approved):

rafaelcoleta:
@leonardokl should we have a unit test to test and document this scenario?

ggdaltoso (approved):

ggdaltoso:
@rafaelcoleta I don't think we should have, at least in this PR.

We can discuss more the needs of what you are pointing and then open a new pull request with these changes. Does it make sense to you?

leonardokl:
@rafaelcoleta We can look for some way to guarantee that all svgs are customizable, through some unit test or a lint. But I think this can be done in another PR.

ggdaltoso:
> @rafaelcoleta We can look for some way to guarantee that all svgs are customizable, through some unit test or a lint. But I think this can be done in another PR.
> I think this scenario is documented at https://gympass.github.io/yoga/components/icon#usage-with-custom-size-fill-and-stroke.

There are two scenarios in this case:
 - the `Icon` from `@gympass/yoga` and
 - all SVGs icons from `@gympass/yoga-icons`

In your PR you're just modifying an SVG icon that must not have a predefined color into it. And we don't have (and I don't know for sure if it must have) tests in this package.

We already have [tests for the Icon](https://github.com/Gympass/yoga/blob/master/packages/yoga/src/Icon/web/Icon.test.jsx#L26), from `@gympass/yoga` asserting the `fill` prop and if it is not enough, we can discuss it in another pull request.

PR #322, opened 2021-08-25
Fix Result component

coelhucas:
This attempt to fix some errors on Result components.
- Many iteration renders without key for children
- Multiple wrong prop types

**⚠️  Although this PR already solve them, I'm not comfortable with what we have now, specially the `JSON.stringify(props)` for keys. I'm opening this to receive suggestions on how to improve it.**

https://user-images.githubusercontent.com/28108272/130797069-137ca75e-23b7-4722-8d78-0e167c174601.mp4


alycecristines (approved):

thiagogympass:
Why don't use `index` as `key` instead `JSON.stringify(props)`?

coelhucas:
> Why don't use `index` as `key` instead `JSON.stringify(props)`?

[no-array-index-key](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md)

packages/yoga/src/Result/native/Tags.jsx
```diff
@@ -25,7 +25,7 @@ const Wrapper = styled.ScrollView`
 const ResultTags = ({ items }) => (
   <Wrapper horizontal showsHorizontalScrollIndicator={false}>
     {items.map(({ ...props }) => (
```
willianpossa, line 27: What is returned in that `...props`?

Don't exist any unique value to use as key in `<TagStyled />?

willianpossa: We can use the `variation` value as key, since that we have only one variant in component. right?

thiagogympass:
> > Why don't use `index` as `key` instead `JSON.stringify(props)`?
> 
> [no-array-index-key](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md)

`no-array-index-key` is a good practice, is not wrong. I think is better and more performative than `JSON.stringify(props)`. `array-index-key` is good to avoid when you remove itens from list, I think is not the case, right?

packages/yoga/src/Result/native/Tags.jsx
coelhucas: cc/ @alycecristines

coelhucas:
> > > Why don't use `index` as `key` instead `JSON.stringify(props)`?
> > 
> > 
> > [no-array-index-key](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md)
> 
> `no-array-index-key` is a good practice, is not wrong. I think is better and more performative than `JSON.stringify(props)`. `array-index-key` is good to avoid when you remove itens from list, I think is not the case, right?

Makes sense. I assume I'm not sure because I had no idea of how is this component, I was just looking at this code. But it seems that its usage really don't expect to change these items

PR #327, opened 2021-08-27
Create list selectable item 

nypacheco:
- Add `small` prop to `List.Item`
- Add `onPress` to `List.Item` that will use `TouchableHighlight` *(from react-native)* when the prop exists.

⚠️ *waiting android screenshots*.

| android |ios |
| ----| ----    |
|listItemAndroid|listItemIos|


marcosricardo (approved):

packages/yoga/src/List/native/ListItem.jsx
```diff
@@ -20,26 +22,55 @@ const StyledView = styled.View(
...
 ListItem.propTypes = {
+  small: PropTypes.bool,
   divided: PropTypes.bool,
+  onPress: PropTypes.func,
 };
 
 ListItem.defaultProps = {
+  small: false,
   divided: true,
+  onPress: null,
```
ggdaltoso, line 71:
```suggestion
  onPress: undefined,
```

just because 
```jsx
typeof null !== typeof(() => {})
```

evilamaior (approved):

packages/yoga/src/List/native/ListItem.jsx
```diff
@@ -20,26 +22,55 @@ const StyledView = styled.View(
...
+            ${list.listItem.padding.left}px;
+        `
+  }
   `,
 );
 
-const ListItem = ({ divided, ...rest }) => (
-  <StyledView divided={divided} {...rest} />
-);
+const ListItem = ({ theme, small, divided, onPress, ...rest }) => {
```
alan-oliv, line 45: hey Yuki, we have a non-documented (sorry sorry sorry, my fault) convention that components use
we agreed to destructure the theme to the declared component level, so we avoid a lot of dot property accessors
what do you think?

alan-oliv (approved):
Impressive contrib! thanks for that

packages/yoga/src/List/native/ListItem.jsx
nypacheco: Done!

packages/yoga/src/List/native/ListItem.jsx
nypacheco: Yes, sure! Updated!

ggdaltoso (approved):
Beautifull! ✨

PR #335, opened 2021-09-08
Add system support for Icon

mmartinsoliveira:
# ✨ Add system support for Icon

image

### Changes 
- Add system props

### Checklist

- [x] Unit tests
- [ ] E2E tests
- [ ] Stories

Co-authors: @ggdaltoso


alan-oliv (approved):
Cool! great contrib!

I noticed that the stroke prop seems to affect the line thickness and it's making the ico component look weird (when it should only change the color), but it's not related to this PR

I'll open a discussion so we can talk about it

ggdaltoso (approved):
proud

coelhucas (approved):

packages/common/src/resolve.js
```diff
@@ -0,0 +1,12 @@
+function resolve(obj, path) {
+  try {
+    return path
+      .split('.')
+      .reduce((prev, curr) => prev && prev[curr], obj)
+      .toString();
```
leonardokl, line 6: Why do we need to convert it to a string?

leonardokl (approved):

packages/common/src/resolve.js
ggdaltoso: We're enforcing `path` to be an object property, not an array position

ggdaltoso: I've replaced the entire resolve for [`lodash.get`](https://lodash.com/docs/4.17.15#get)

But we could also use [`just-safe-get`](https://www.npmjs.com/package/just-safe-get), as suggested by @leonardokl 

What do you think @tcK1?

packages/system/src/theme.js
```diff
@@ -1,5 +1,5 @@
 import { css } from 'styled-components';
-import { resolve } from '@gympass/yoga-common';
+import { get } from 'lodash';
```
tcK1, line 2: I think this should be like this instead:

```suggestion
import get from 'lodash.get';
```

Using it from `lodash` means that we expect the package `lodash` as a dependency right? This only works because one of our dependencies has is as a dependency (dependency inception 😂 ).

tcK1 (dismissed):

packages/common/src/resolve.js
tcK1: I really liked this `just-safe-get` package! Maybe we could bench them and see if there's a difference (Since these functions will probably be called sometimes).

But all in all, both choices seems nice!

packages/system/src/theme.js
ggdaltoso: _blaming VSCode auto-import_

leonardokl (approved):

PR #341, opened 2021-09-17
Add system support for Text

ericcleao:
# ✨ Add system support for Text

### Changes 
- Add system props

### Checklist

- [x] Unit tests
- [ ] E2E tests
- [ ] Stories

Co-authors: @ggdaltoso

### DIFF
--- a/packages/doc/content/components/components/text/default-native.mdx
+++ b/packages/doc/content/components/components/text/default-native.mdx
@@ -1,3 +1,5 @@
+import { Link as GatsbyLink } from 'gatsby';
+
 ### Usage
 
 ```javascript type=expo
@@ -34,5 +36,7 @@
 ```
 
 ### Props
+
+The Text component also has support for system props as you can see more details <GatsbyLink to="/system/getting-started">our system</GatsbyLink>
 
 <PropsTable component="Text" platform="native" />
--- a/packages/doc/content/components/components/text/default-web.mdx
+++ b/packages/doc/content/components/components/text/default-web.mdx
@@ -1,3 +1,5 @@
+import { Link as GatsbyLink } from 'gatsby';
+
 ### Simple Usage
 
 ```javascript reactLive=true
@@ -55,5 +57,7 @@ provided by the `<ThemeProvider />`, if you are using Server Side Rendening
 (SSR) you need to import the `<FontLoader />` in your `<Head>`.
 
 ### Props
+
+The Text component also has support for system props as you can see more details <GatsbyLink to="/system/getting-started">our system</GatsbyLink>
 
 <PropsTable component="Text" platform="web" />
--- a/packages/yoga/src/Snackbar/web/Snackbar.jsx
+++ b/packages/yoga/src/Snackbar/web/Snackbar.jsx
@@ -142,9 +142,9 @@ const Snackbar = ({
           </Box>
         )}
 
-        <Box as={Text.Small} flex={1} mr="small" numberOfLines={2}>
+        <Text.Small flex={1} mr="small" numberOfLines={2}>
           {message}
-        </Box>
+        </Text.Small>
 
         <ActionsWrapper>
           {onAction && actionLabel && (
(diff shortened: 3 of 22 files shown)


packages/doc/content/components/components/text/default-web.mdx
```diff
@@ -56,4 +58,6 @@ provided by the `<ThemeProvider />`, if you are using Server Side Rendening
 
 ### Props
 
+The Text component also has support for system props as you can see more details <GatsbyLink to="/system/getting-started">our system</GatsbyLink>
+
```
coelhucas, line 62: Nit
```suggestion
Text component also has support for system props as you can see <GatsbyLink to="/system/getting-started">here</GatsbyLink>

```

packages/doc/content/components/components/text/default-native.mdx
```diff
@@ -35,4 +37,6 @@
 
 ### Props
 
+The Text component also has support for system props as you can see more details <GatsbyLink to="/system/getting-started">our system</GatsbyLink>
+
```
coelhucas, line 41: Nit
```suggestion
Text component also has support for system props as you can see <GatsbyLink to="/system/getting-started">here</GatsbyLink>

```

coelhucas (commented):
Nice job with te `deprecated` abstraction 😉 
You dropped all of the code I did before, though 💔  but ok, no hard feelings 🥺

packages/doc/content/components/components/text/default-web.mdx
ericcleao: Updated 👍

packages/doc/content/components/components/text/default-native.mdx
ericcleao: Updated 👍

coelhucas (approved):
LGTM

ggdaltoso (approved):
That's a really good contribution!!

I was waiting for it!

packages/yoga/src/shared/propTypes/deprecated.js
```diff
@@ -0,0 +1,23 @@
...
+
+    return propType(props, propName, componentName, ...rest);
+  };
+}
+
+function resetWarned() {
+  warned = {};
+}
+
+deprecated.resetWarned = resetWarned;
```
alan-oliv, line 23: Hey man! Nice use of [deprecated prop-type](https://reactjs.org/warnings/dont-call-proptypes.html#fixing-the-false-positive-in-third-party-proptypes)!

I was just thinking here, what's the use for the reset on the yoga context?
can we simplify this implementation?

Like, if we remove the reset there's no need for the `warned` let outside the deprecated function
we could also use the export convention of the other files, creating a `const` and exporting at the end of the file (not really necessary, but just to keep code consistency)


another thing I was looking over here, is that, react-proptypes uses `warning` [as a dependency](https://github.com/react-bootstrap/react-prop-types/blob/master/package.json#L33) but yoga doesn't, is this really working?

ReactJS package is using `console.warn` for [deprecation warning since 15.6.0](https://github.com/reactjs/reactjs.org/blob/a4d4893bd3c56f4f7f821c46ed512eac6c3c21b9/content/blog/2017-06-13-react-v15.6.0.md), can we use it too?

what do you think?

ericcleao: > ReactJS package is using console.warn for deprecation warning since 15.6.0, can we use it too?

Agree, updated to use the `console.warn` itself.
And also removed the unnecessary reset function.

mmartinsoliveira (approved):

packages/yoga/src/shared/propTypes/deprecated.js
alan-oliv: 🚀

alan-oliv (approved):

packages/yoga/src/Snackbar/web/Snackbar.jsx
```diff
@@ -142,9 +142,9 @@ const Snackbar = ({
           </Box>
         )}
 
-        <Box as={Text.Small} flex={1} mr="small" numberOfLines={2}>
+        <Text.Small flex={1} mr="small" numberOfLines={2}>
```
Wendlereis, line 145: Pleasant!!

Wendlereis (approved):

ggdaltoso:
@all-contributors please add @ericcleao for code

PR #344, opened 2021-09-23
Implement Theme Generator function `createTheme` with consumption by `<ThemeProvider>`

coelhucas:
## Theme what?
You read it right. This new API enables any Yoga consumer to inject their very own themes.

### How then?
```jsx
import { 
  createTheme,
  Button,
  ThemeProvider,
} from '@gympass/yoga';

const myTheme = createTheme(tokens => ({
  colors: {
    primary: tokens.colors.deepPurple,
  },
});

const App = () => (
  <ThemeProvider theme={myTheme}>
    <ApplicationComponent />
    <Button>This button is already going to be themed!</Button>
  </ThemeProvider>
);
```

image

Just like that. You can thank me later. For now, suggestions for improvements, change requests and etc are highly appreciated. 

------
:warning:  You can notice that `createTheme` just returns what it receives, which means that this would work:
```jsx
const App = () => (
  <ThemeProvider theme={(t) => { colors: { primary: t.colors.hope } }}>
    <Button>I'm also themed</Button>
  </ThemeProvider>
);
```

Besides that, I'd enforce to **not do** it this way, since we're not validating the received param and if for whatever reason we make more validations and updates in `createTheme`, your usage will likely to broke in new versions more than expected.


Wendlereis (approved):

ggdaltoso (approved):
beautiful documentation and JSDoc implementation as well

This PR moves Yoga to a new level

Thanks, @coelhucas

mmartinsoliveira (approved):

CONTRIBUTING.md
```diff
@@ -64,15 +64,9 @@ packages
...
+### Theming
 
-Currently, we have 3 predefined themes `Corporate`, `EndUser` and `Gyms`, all
-of them follows the `BaseTheme` object.
-
-Every component has its own `theme` file. You can find it in its own folder.
-
-When building a new component, make sure to add values like paddings, colors,
-margins, to the component theme file.
+Yoga exposes a `createTheme` function in order to generate new themes, you can refer to [Theme Generator](https://gympass.github.io/yoga/components/theming/themeGenerator) in order to see how to create custom themes.
```
LidiaYoshino, line 69:
```
Every component has its own `theme` file. You can find it in its own folder.

When building a new component, make sure to add values like paddings, colors,
margins, to the component theme file.
```
Is this not true anymore?

If it still is, I'd keep this info

coelhucas: What a pair of hawk eyes! Thank you a lot, indeed I mistakenly removed more than needed

packages/yoga/src/Theme/theme/theme.js
```diff
@@ -1,7 +1,24 @@
...
+  const components = {};
+  Object.entries(componentThemes).forEach(([names, themed]) => {
+    const [, name] = names.split('$');
+    components[name.toLowerCase()] = themed({
+      ...tokens,
+      colors,
+      baseFont,
+      baseFontSize,
+    });
+  });
```
tcK1: This is a personal opinion:

```suggestion
  const components = Object.entries(componentThemes).reduce((acc, [names, themed]) => {
    const [, name] = names.split('$');
    acc[name.toLowerCase()] = themed({
      ...tokens,
      colors,
      baseFont,
      baseFontSize,
    });
  }, {});
```

packages/yoga/src/Theme/Provider/Provider.jsx
```diff
@@ -4,14 +4,25 @@ import PropTypes from 'prop-types';
...
+ * @param {function} themeParameters.theme returns a custom theme object to replace existing tokens
+ * @returns {Object.<string, string | number>} theme object with tokens sent to styled-components' <ThemeProvider />
+ */
+const getTheme = ({ locale, theme }) => {
   const token = tokens[locale] || tokens.default;
-  return yogaTheme(token);
+  const defaultTheme = yogaTheme(token);
+  const customTheme = theme ? yogaTheme(defaultTheme, theme(token)) : null;
+
+  return customTheme || defaultTheme;
```
tcK1, line 20:
```suggestion
  return theme
    ? yogaTheme(defaultTheme, theme(token))
    : yogaTheme(token);
```

packages/yoga/src/Theme/theme/theme.js
```diff
@@ -1,6 +1,24 @@
...
+
+  Object.entries(componentThemes).forEach(([names, themed]) => {
+    const [, name] = names.split('$');
+    components[name.toLowerCase()] = themed({
+      ...tokens,
+      colors,
+      baseFont,
+      baseFontSize,
+    });
+  });
```
tcK1, line 17: This is a personal opinion:

```suggestion
  const components = Object.entries(componentThemes).reduce((acc, [names, themed]) => {
    const [, name] = names.split('$');
    acc[name.toLowerCase()] = themed({
      ...tokens,
      colors,
      baseFont,
      baseFontSize,
    });
  });
```

tcK1 (changes requested):
We can add validation on the `composeTheme` function, so it will only work with functions generated by the `createTheme` function.

For example:

```js
function createTheme(themeBuilder) {
  const themeBuilderCloned = themeBuilder.bind({});
  themeBuilderCloned.prototype.isFromCreateTheme = true;
  return themeBuilderCloned;
}

const getTheme = ({ locale, theme }) => {
  if (theme && theme.hasOwnProperty('isFromCreateTheme') throw new Error('oops!');
};
```

This is just an example! 😉

packages/yoga/src/Theme/theme/theme.js
coelhucas: Applied this, with a working reducer 🤪
Thanks!

ericcleao (approved):

Falkaniere (approved):

packages/yoga/src/Theme/helpers/themeGenerator/themeGenerator.js
```diff
@@ -0,0 +1,25 @@
+/**
+ * Gets a function that one could sent to \<ThemeProvider> in order to inject a custom thme (or sets of tokens)
+ * @param {themeVars} themeVars - should return an object with a similar scheme of our tokens that will be replaced, can receive a tokens parameter
+ * @returns {themeVars} themeVars - will return the same function which is injected to \<ThemeProvider>
+ */
+function createTheme(themeVars) {
+  if (typeof themeVars !== 'function')
+    throw new TypeError(
+      `Invalid param 'themeVars', expected 'string', received '${typeof themeVars}'.`,
```
tcK1, line 9:
```suggestion
      `Invalid param 'themeVars', expected 'function', received '${typeof themeVars}'.`,
```

packages/yoga/src/Theme/helpers/themeGenerator/themeGenerator.js
```diff
@@ -0,0 +1,25 @@
...
+    );
+
+  const themeBuilder = themeVars;
```
tcK1, line 12: Maybe we should clone the function instead of working with the one we received.

```suggestion
  const themeBuilder = themeVars.bind({});
```

packages/yoga/src/Theme/Provider/Provider.jsx
```diff
@@ -4,15 +4,36 @@ import PropTypes from 'prop-types';
...
+ */
+const getTheme = ({ locale, theme }) => {
+  const isThemeInvalid = theme ? !('isFromThemeBuilder' in theme) : false;
+
+  if (isThemeInvalid) {
+    throw new TypeError(
+      `Invalid 'customTheming' parameter. Expected one generated by createTheme() function.
+    Did you sent a function directly to <ThemeProvider />?
+    
+    const getTheme = ({ locale, theme }) => {
```
tcK1, line 23: I don't think that this error is helpful, since the end-user does not work with this function he doesn't know what `getTheme` means. Maybe we should say something like "_The Provider prop has an invalid type_" or something like that.

What do you think?

tcK1 (changes requested):

packages/yoga/src/Theme/Provider/Provider.jsx
coelhucas: The stack trace would already point to the internal source, I believe that `'customTheming'` isn't clear enough (changing it rn), but providing the engineer more context after the original message is good so one may know where to see the implementation in order to understand it better.

The "helpful" information (the one expected to quickly solve the issue is the first one. Makes sense to you?

cc/ @ggdaltoso

packages/yoga/src/Theme/Provider/Provider.jsx
```diff
@@ -1,18 +1,50 @@
...
+/**
+ * Gets resolved theme to inject in Yoga
+ *
+ * @param {{locale: string, theme: function}} themeParameters information about the theme
+ * @param {string} themeParameters.locale custom locale to get colors from (if it has a different scheme)
+ * @param {function} themeParameters.theme returns a custom theme object to replace existing tokens
+ * @returns {Object.<string, string | number>} theme object with tokens sent to styled-components' <ThemeProvider />
+ */
+const getTheme = ({ locale, theme }) => {
+  const isThemeInvalid = theme ? !('isFromThemeBuilder' in theme) : false;
```
tcK1, line 16: Seems like we have multiple branching only to validate the theme. What do you think of something like this:

```js
const validateAndRunBuilder = (builder, tokens) => {
  if (typeof builder === 'undefined') return false;
  if (!'isFromThemeBuilder' in theme) throw /* ... */;
  
  const customTheme = builder(tokens);
  if (typeof customTheme !== 'object') throw /* ... */;

  return customTheme;
}

const getTheme = ({ theme }) => {
  const token = tokens[locale] || tokens.default;
  const customTheme = validateAndRunBuilder(theme, token);

  /* ... */

  return customTheme ? yogaTheme(token, customTheme) : defaultTheme;
}
```

tcK1 (approved):

PR #345, opened 2021-09-24
Add `padding-line-between-statements` rule

coelhucas:
### About
This gives us less concern while coding and reviewing PRs. Does you prefer to just write the code and let your editor's save fix it? Be our guest.

See [`padding-line-between-statements`](https://eslint.org/docs/rules/padding-line-between-statements).

### New rules
What changes with them?

❌  **Bad** - No lines between declarations and other statements:
```js
const a = 42;
const b = 'coelhucas';
function getInfo() {
  return { a, b };
}
```

✅  **Good** - Blank line between declarations and other statements
```js
const a = 42;
const b = 'coelhucas';

function getInfo() {
  return { a, b };
}
```

❌  **Bad** - No lines between directives and other statements:
```js
"use strict"
const name = 'Lucas';
```

✅  **Good** - Blank line between directives and other statements:
```js
"use strict";

const name = 'Lucas';
```

You got the idea, right?

✅  Also, blank lines between declarations are **OK**:
```js
const sayMyName = 'Walter White';

const sum = (a, b) => a + b;
```

✅  And **NO** line breaks between directives are also **OK**:
```js
"use strict";
"use asm";
```

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -21,6 +21,17 @@ module.exports = {
     'import/no-extraneous-dependencies': 0,
     'no-plusplus': 0,
     'no-restricted-globals': 0,
+    'padding-line-between-statements': [
+      2,
+      { blankLine: 'always', prev: ['const', 'var', 'let'], next: '*' },
+      {
+        blankLine: 'any',
+        prev: ['const', 'var', 'let'],
+        next: ['const', 'var', 'let'],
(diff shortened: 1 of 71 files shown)


ggdaltoso (approved):
🧹

ericcleao (approved):

PR #346, opened 2021-09-24
Add text align css property into typography module (Yoga system)

lucasfernandesbr:
# ✨ Add text align prop to system

image

## Motivation

Keeping the work in `@gympass/system`, we're adding the text align property to typography's module:

- `textAlign`, `ta`

Those updates will be reflected into components that are currently using the `yoga-system` module.

### Checklist

- [x] Unit tests

### DIFF
--- a/packages/system/src/typography.test.js
+++ b/packages/system/src/typography.test.js
@@ -170,20 +170,16 @@ describe('Web and iOS', () => {
     });
 
     describe('textAlign', () => {
-      it('Should return values for text align prop', () => {
+      it('Should return values for textAlign prop', () => {
         const expectedTextAlign = css({ textAlign: 'center' });
 
-        const useSystemTextAlign = textAlign({ textAlign: 'center' });
+        const ta1 = textAlign({ textAlign: 'center' });
+        const ta2 = textAlign({ ta: 'center' });
 
-        expect(useSystemTextAlign).toStrictEqual(expectedTextAlign);
-      });
+        expect(ta1).toStrictEqual(ta2);
 
-      it('Should return values for text align prop when using the abbreviation version', () => {
-        const useSystemTextAlign = css({ textAlign: 'center' });
-
-        const textAlignProperty = textAlign({ ta: 'center' });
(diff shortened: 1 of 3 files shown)


mmartinsoliveira (approved):

andreloureiro (approved):

Falkaniere (approved):

ericcleao (approved):

packages/system/src/typography.test.js
```diff
@@ -167,6 +168,24 @@ describe('Web and iOS', () => {
...
+        expect(useSystemTextAlign).toStrictEqual(expectedTextAlign);
+      });
+
+      it('Should return values for text align prop when using the abbreviation version', () => {
+        const useSystemTextAlign = css({ textAlign: 'center' });
+
+        const textAlignProperty = textAlign({ ta: 'center' });
+
+        expect(textAlignProperty).toStrictEqual(useSystemTextAlign);
+      });
```
ggdaltoso, line 187: I think we could make just one scenario testing both props. It would also assert if `textAlign` === `ta`

```suggestion
            it('Should return values for textAlign prop', () => {
        const expectedTextAlign = css({ textAlign: 'center' });

        const ta1 = textAlign({ textAlign: 'center' });
        const ta2 = textAlign({ ta: 'center' });

        expect(ta1).toStrictEqual(ta2);

        const textAlignOptions = [ta1, ta2];
        textAlignOptions.map(ta => expect(ta).toStrictEqual(expectedTextAlign));
      });
```

What do you think?

ggdaltoso (changes requested):

packages/system/src/typography.test.js
lucasfernandesbr: That's awesome! It gives a better visualization of the test. I'm commiting the suggestion 🚀 !!

ggdaltoso (approved):

coelhucas:
@all-contributors please add @lucasfernandesbr for code

PR #349, opened 2021-09-29
✨ New component: Button Icon

marcosricardo:
We created the component **Button Icon** in native and web

[Figma prototype](https://www.figma.com/file/wn9grDi9mM29Wt7LmQjol7/Yoga-Design-System-7.1?node-id=12958%3A11291)

**Native:**

Screen Shot 2021-09-29 at 10 39 02

**Web:**

Screen Shot 2021-09-29 at 10 43 29


packages/yoga/src/Button/native/Icon.jsx
```diff
@@ -0,0 +1,141 @@
...
+    pressed,
+  }) => {
+    const state = secondary ? 'secondary' : 'primary';
+
+    return `
+      display:flex;
+      align-items: center;
+      justify-content: center;
+      
+      border-radius: ${large ? 24 : 12}px;
```
ggdaltoso, line 27: Should we use just `radii.circle` from theme instead of these `12|24` values?

packages/yoga/src/Button/native/Icon.jsx
```diff
@@ -0,0 +1,141 @@
...
+    >
+      <Icon
+        as={icon}
+        fill={textColor}
+        width={
+          large ? button.types.icon.svg.large : button.types.icon.svg.default
+        }
+        height={
+          large ? button.types.icon.svg.large : button.types.icon.svg.default
+        }
```
ggdaltoso, line 114:
```suggestion
        size={
          large ? button.types.icon.svg.large : button.types.icon.svg.default
        }
```

packages/yoga/src/Button/native/Icon.jsx
```diff
@@ -0,0 +1,141 @@
...
+    disabled,
+    inverted,
+    pressed,
+  }) => {
+    const state = secondary ? 'secondary' : 'primary';
+
+    return `
+      display:flex;
+      align-items: center;
+      justify-content: center;
```
ggdaltoso, line 25: We usually add the CSS properties which aren't related to the theme or props to be outside the callback

packages/yoga/src/Button/native/Icon.jsx
```diff
@@ -0,0 +1,141 @@
...
+      };
+     
+      width: ${
+        large ? button.types.icon.size.large : button.types.icon.size.default
+      }px;
+      height: ${
+        large ? button.types.icon.size.large : button.types.icon.size.default
+      }px;
+
+
```
ggdaltoso, line 40:
```suggestion
```

packages/yoga/src/Button/native/Icon.jsx
```diff
@@ -0,0 +1,141 @@
...
+          ? `
+          background-color: ${button.types.contained.font.default.color};
+        `
+          : ''
+      }
+    `;
+  }}
+`;
+
+const ButtonIcon = ({
```
ggdaltoso, line 68: Should we export ButtonIcon, on Native, using `forwardRef` too?
You guys are doing it just for Web

packages/yoga/src/Button/web/Icon.jsx
```diff
@@ -0,0 +1,125 @@
...
+            `
+          : ''
+      }
+    `;
+  }}
+`;
+
+const ButtonIcon = forwardRef(({ icon: Icon, ...props }, ref) => (
+  <IconStyled ref={ref} {...props}>
+    <Icon />
```
ggdaltoso, line 102: @ericcleao Do you think we should use `Icon`, from Yoga, to handle this component?
It would change how the `Button.Icon.theme` has been set. 

Do you think it makes sense and is a valid effort?

ggdaltoso (changes requested):

coelhucas:
You folks can hate rebase, but it saved the merge issue 😂

packages/yoga/src/Button/native/Icon.jsx
hesugiyama: commited :)

packages/yoga/src/Button/native/Icon.jsx
hesugiyama: commited :)

packages/yoga/src/Button/web/Icon.jsx
ericcleao: I believe it is better to use the Icon whenever we have an Icon at other siblings components because we gain with all the other features that are already implemented at the Icon component like `size`, `fill`, `stroke`, `spacing` ...

ericcleao: In this context, IconStyled is a Button that contains a "dummy" SVG icon, but if we use the Icon we could define the size and margins at itself, together with all the other design rules that could come over at the future

packages/yoga/src/Button/native/Icon.jsx
hesugiyama: commited :)

packages/yoga/src/Button/web/Icon.jsx
ericcleao: @marcosricardo and @ggdaltoso, do you think that the gain is good enough to explain the effort that will take to changing the approach?

packages/yoga/src/Button/native/Icon.jsx
hesugiyama: commited :)

packages/yoga/src/Button/native/Icon.jsx
```diff
@@ -0,0 +1,137 @@
...
+  );
+};
+
+ButtonIcon.propTypes = {
+  large: bool,
+  disabled: bool,
+  pressed: bool,
+  inverted: bool,
+  secondary: bool,
+  /** an Icon from yoga-icons package */
```
coelhucas, line 122: This comment could be safely removed as it's kinda misleading, a we accept any Icon.
```suggestion
```

packages/yoga/src/Button/native/Icon.jsx
```diff
@@ -0,0 +1,137 @@
...
+  icon: oneOfType([node, func]),
+};
+
+ButtonIcon.defaultProps = {
+  large: undefined,
```
coelhucas, line 127: Since it is a boolean property, we can use `false` as default value which will be as falsy as undefined.
```suggestion
  large: false,
```

packages/yoga/src/Button/web/Icon.jsx
coelhucas: > but if we use the Icon we could define the size and margins at itself, together with all the other design rules that could come over at the future

I don't see this as a gain at all. I believe that one shouldn't be able to customize the Icon inside, since one of the premises of the design system is to be somehow opinionated on how one can use its elements. Letting the internal Icon customizable could lead other teams to make UI that shouldn't be possible since they don't follow Yoga principles.

Unlike Icon within Button, the container itself (our Button), allows system usage as the default button. An Icon container in a button shouldn't (IMO) be as "customizable" as an individual Icon component

Does this makes sense for you folks?

packages/yoga/src/Button/web/Icon.jsx
```diff
@@ -0,0 +1,129 @@
...
+  </IconStyled>
+));
+
+ButtonIcon.propTypes = {
+  large: bool,
+  disabled: bool,
+  secondary: bool,
+  inverted: bool,
+  /** an Icon from yoga-icons package */
```
coelhucas, line 115: Same as in native
```suggestion
```

packages/yoga/src/Button/web/Icon.jsx
```diff
@@ -0,0 +1,129 @@
...
+  icon: oneOfType([node, func]),
+};
+
+ButtonIcon.defaultProps = {
+  large: undefined,
```
coelhucas, line 120: Same as native
```suggestion
  large: false,
```

packages/yoga/src/Button/web/Icon.jsx
```diff
@@ -0,0 +1,129 @@
...
+  disabled: undefined,
+  secondary: undefined,
+  inverted: undefined,
```
coelhucas, line 123: Same as native
```suggestion
  disabled: false,
  secondary: false,
  inverted: false,
```

coelhucas (changes requested):
My messages regarding prop types are more opinatives than real change requests, so please get feedback from other engineers before applying them.

packages/yoga/src/Button/web/Icon.jsx
ericcleao: I don't know if I write it wrong but what I meant was that the yoga developers could customize the Icon itself using the Icon props instead of using the svg inside the StyledIcon, that way we have more control and the gain to have all the Icon features to be used at the ButtonIcon.

coelhucas: Sorry, @ericcleao I misunderstood your suggestion. I definitely agree with this suggestion, this refactor will make the code more accurate with how we expected someone to use Yoga with system.

mmartinsoliveira (approved):

hesugiyama (approved):

coelhucas (approved):
LGTM! 🎈

packages/yoga/src/Button/Button.theme.js
```diff
@@ -124,6 +124,20 @@ const Button = ({
         bottom: spacing.xxsmall,
       },
     },
+    icon: {
+      size: {
+        default: 24,
```
ericcleao, line 129: Why do not use `spacing.large` or `'large'` (depending on the context) instead of `24`, it's also not necessary once we are defining the icon size and its margin.

ggdaltoso:
This is what we have on our documentation now

image

Also, this is what we have on mobile as well

image

coelhucas:
@all-contributors pleas add @coelhucas for review (yeah I'm self promoting me 😂)

coelhucas:
@all-contributors please add @marcosricardo for code

PR #350, opened 2021-09-29
Fix at Tag.Informative

ericcleao:
## Changes:

- Set variant props as optional;
- Fix the color used at the default variant;
- Start to use the `Icon` component at the web `Tag`;
- Refactor some details using the system that is integrated at the `Icon`.

## Doc

image (1)

### DIFF
--- a/packages/yoga/src/Tag/native/Informative.jsx
+++ b/packages/yoga/src/Tag/native/Informative.jsx
@@ -4,6 +4,7 @@ import { func, oneOf, node, bool } from 'prop-types';
 import { margins } from '@gympass/yoga-system';
 
 import { StyledTag, StyledText } from './Tag';
+import Icon from '../../Icon';
 
 const Informative = styled(StyledTag)`
   ${({
@@ -54,12 +55,10 @@ const StyledTextInformative = styled(StyledText)`
 /** Tags should be keywords to categorize or organize an item. */
 const TagInformative = ({
   children,
-  icon: Icon,
+  icon,
   theme: {
     yoga: {
-      colors: {
-        text: { primary },
-      },
+      colors: { text },
       components: { tag },
     },
   },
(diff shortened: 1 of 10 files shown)


packages/yoga/src/Tag/native/Informative.jsx
```diff
@@ -81,14 +82,15 @@ const TagInformative = ({
 
 TagInformative.propTypes = {
   /** style the tag following the theme (success, informative, attention) */
-  variant: oneOf(['success', 'informative', 'attention']).isRequired,
+  variant: oneOf(['', 'success', 'informative', 'attention']),
```
coelhucas, line 85: What do you think about having a `default` variant instead?

```suggestion
  variant: oneOf(['default', 'success', 'informative', 'attention']),
```
Then below
```diff
TagInformative.defaultProps = {
--  variant: '',
++  variant: 'default',
```
Same suggestion for web

coelhucas (commented):
I have mixed feelings when it comes to using the logical and to conditional renderings, I'm leaning towards ternaries instead, but you can also convert icon prop to boolean with `!!icon`. Up to you, this isn't necessarily a change request.

[Reference on this topic](https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx).

ggdaltoso:
We have some misconceptions between our designers and it would be better to hold this PR closed until they decide what they want to do within this default Tag

ggdaltoso:
@ericcleao 

There is some good things on this PR we could take advantage of.

What do you think about redoing the changes adding the gray Tag and just ship the refactors?
In addition to that, we could add a better default prop to Tag, preventing of been use wrongly.

ericcleao:
@ggdaltoso I believe it's easier if we just close this PR and use the interesting part that involves the Icon at the Tag component at other PR, and afterward, we update the tag default color if it's needed.

ericcleao:
New PR updating the icon approach at the Tag: https://github.com/Gympass/yoga/pull/355

PR #355, opened 2021-10-04
Start to use Icon Component inside the Tag instead of SVG

ericcleao:
## Description

Changed the icon usage at the Tag component, start using the Yoga Icon component instead of rendering the icon as an SVG.

As you can see in the below doc screenshot the icon `marginRight` is set as `xxxsmall` and the SVG style has the `margin-right` as `4px`.

## Doc screenshot

image

### DIFF
--- a/packages/yoga/src/Tag/Tag.theme.js
+++ b/packages/yoga/src/Tag/Tag.theme.js
@@ -5,7 +5,7 @@ const Tag = ({ spacing, fontSizes, fontWeights, radii, borders }) => ({
       small: spacing.xsmall,
     },
     margin: {
-      right: spacing.xxxsmall,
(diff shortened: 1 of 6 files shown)


ggdaltoso (approved):

packages/yoga/src/Tag/Tag.theme.js
```diff
@@ -5,7 +5,7 @@ const Tag = ({ spacing, fontSizes, fontWeights, radii, borders }) => ({
       small: spacing.xsmall,
     },
     margin: {
-      right: spacing.xxxsmall,
+      right: 'xxxsmall',
```
caioalexandrebr, line 8: ❤️

caioalexandrebr (approved):

Issue #343, opened 2021-09-23
[BUG] Navigation Bar overlaying mobile menu/sidebar


caioalexandrebr:
**Describe the bug**
Currently the mobile menu/sidebar is cutting the presentation of items positioned at the bottom of the list. Chrome NavigationBar overlaps the end of the listing.

https://user-images.githubusercontent.com/31045534/134536445-c2d2a72a-7f3b-49d9-9f09-b2de3dd175d7.mp4

**To Reproduce**
1. Go to https://gympass.github.io/yoga/components/getting-started/
2. Scroll to the bottom of the page.
3. In the menu mobile/sidebar, the last items listed are off the screen.

**Expected behavior**
It will be possible to view all items listed in the menu mobile/sidebar.

**Smartphone:**
 - iPhone 11
 - OS: 14.7
 - Chrome

coelhucas:
Are you willing to pick up this issue?

ggdaltoso:
Seems to be an easy fix. You just need to update a CSS unit from `vh` to `%` [here](https://github.com/Gympass/yoga/blob/master/packages/doc/src/components/Navigation/Navigation.jsx#L40)

```diff
-      height: calc(100vh - 70px);
+      height: calc(100% - 70px);
```

caioalexandrebr:
Hi guys, thank you so much for your collaboration! 🚀 

I would love to fix this problem.

ty @coelhucas @ggdaltoso ❤️

coelhucas:
As we discussed internally, it's being addressed in #356.

coelhucas:
Fixed in #356 

| iOS | Android | Desktop |
| -- | -- | -- |
| ![image0](https://user-images.githubusercontent.com/28108272/136259115-7aae4848-33e4-4fe2-a0c9-f5a217af5e9d.png) | ![Screenshot_20211006-150331_Firefox Nightly](https://user-images.githubusercontent.com/28108272/136259137-b230f82b-63c5-47b4-926e-3decbc17a0eb.jpg) | ![image](https://user-images.githubusercontent.com/28108272/136259479-74974e5b-e675-49ea-b14d-f40ca457ae55.png) |


An interesting behaviour that I noticed, though, is that the order changed in my iPhone 🤔. But in my tablet, desktop and @caioalexandrebr's iPhone it keeps the same order. cc/ @ggdaltoso

coelhucas:
[...]

PR #356, opened 2021-10-05
[DOCS: A11y | Navigation] Make it accessible through keyboard

coelhucas:
> "That's one small step for man, one giant leap for mankind."

This PR enables keyboard navigation through our sidebar. ✨ This brings us:
- Improved semantics;
- More accessibility (still room to improve in our documentation in general);
- Room and example for future improvements regarding a11y;
- Fix for #343;

It also apply some considerable refactoring into our Navigation component inside docs
This is the current status of the navigation using macOS' VoiceOver (turn sound on)

https://user-images.githubusercontent.com/28108272/136059501-ce994485-a55a-462d-8eed-7c328d9fb040.mp4

Some decisions made:
- Replace px font-size with rem, since rem adapts based on user's font size configuration;
- Add a little transition and rotation to the Collapsible arrow, illustrating its current state
- Use an if statement to conditionally render collapsible of link, since it improved a lot this code readability
- Move some booleans to declarations outside, also improving JSX readability

| Before | After | Focused (after) |
| --- | --- | --- |
| image | image | image |

There is a slightly difference between before and after due to the change to rem.

### DIFF
--- a/packages/doc/src/components/Navigation/Navigation.jsx
+++ b/packages/doc/src/components/Navigation/Navigation.jsx
@@ -239,7 +239,7 @@ const Navigation = ({ items, toggleMenu, opened, prefix }) => {
 
   return (
     <Wrapper opened={opened}>
-      <Nav aria-label="Contents navigation">
+      <Nav>
         <List tree={tree} toggleMenu={toggleMenu} prefix={prefix} />
       </Nav>
     </Wrapper>
--- a/packages/doc/src/components/Navigation/Navigation.jsx
+++ b/packages/doc/src/components/Navigation/Navigation.jsx
@@ -138,7 +138,7 @@ const ListItem = ({
 
   const { pathname } = window.location;
   const linkPath = prefix ? `/yoga${filteredUrl}` : filteredUrl;
-  const isActive = pathname === linkPath;
+  const isActive = window && pathname.replace(/\/$/, '') === linkPath;
 
   const onNavigate = () => {
     if (filteredUrl !== pathname) {
--- a/packages/doc/src/components/Navigation/Navigation.jsx
+++ b/packages/doc/src/components/Navigation/Navigation.jsx
@@ -37,7 +37,7 @@ const Wrapper = styled.aside`
       left: 0;
 
       width: 100%;
-      height: calc(100vh - 70px);
+      height: calc(100% - 70px);
 
       overflow: auto;
       z-index: 3;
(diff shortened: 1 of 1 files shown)


packages/doc/src/components/Navigation/Navigation.jsx
```diff
@@ -57,35 +57,22 @@ const Nav = styled.div`
...
-    :hover {
-      color: ${primary};
-    }
-  `};
+const ArrowIcon = styled(Arrow)`
+  ${({ isOpen }) => `
+    width: 0.6rem;
+    transition: all 200ms ease-out;
+    transform: rotate(${isOpen ? 180 : 0}deg);
+  `}
```
ggdaltoso, line 75:
```suggestion
  width: 0.6rem;
  transition: all 200ms ease-out;
  
  ${({ isOpen }) => `
    transform: rotate(${isOpen ? 180 : 0}deg);
  `}
```

packages/doc/src/components/Navigation/Navigation.jsx
```diff
@@ -109,23 +96,44 @@ const StyledListItem = styled.li`
...
+  text-decoration: none;
+  font-weight: 300;
+  align-items: center;
+  width: 100%;
+  background: none;
+  border: none;
+  transition: all 200ms ease-out;
+  color: ${hexToRgb(text.secondary, 0.75)};
+  text-indent: calc(15px * ${level});
+  outline-offset: 2px;
```
ggdaltoso, line 124: Can you move all CSS properties which doesn't make use of the callback function outside of it?

ggdaltoso (changes requested):
Thanks for taking care of this. It is very important!!!

Also, what do you think about adding a fix solving #343?

mmartinsoliveira (approved):

packages/doc/src/components/Navigation/Navigation.jsx
```diff
@@ -240,7 +239,7 @@ const Navigation = ({ items, toggleMenu, opened, prefix }) => {
 
   return (
     <Wrapper opened={opened}>
-      <Nav>
+      <Nav aria-label="Contents navigation">
```
LidiaYoshino, line 242:
```suggestion
      <Nav aria-label="Navigation content">
```

coelhucas: As we discussed internally, I'm actually removing this whole aria-* attribute since nav itself already includes it and is accessible by default 🌻

PR #366, opened 2021-10-08
✨ Add new props gap to system

caioalexandrebr:
## Motivation

Keeping the work in `@gympass/system`, we're adding the follow new props:

- `gap`

Those updates will be reflected into our Box component as it makes use of the `system` module

### Checklist

- [x] Unit tests

### DIFF
--- a/packages/doc/content/system/system/flex.mdx
+++ b/packages/doc/content/system/system/flex.mdx
@@ -199,7 +199,7 @@ The `flexes` module has all the listed above
         <inlineCode>gap</inlineCode>
       </td>
       <td>
-        <inlineCode>gap</inlineCode>, <inlineCode>g</inlineCode>
+        <inlineCode>gap</inlineCode>
       </td>
       <td>
         <inlineCode>gap</inlineCode>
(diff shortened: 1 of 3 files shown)


coelhucas:
I'd like to bring further discussion to this, since I believe we may be getting pretty out of hand with some abbreviations (such as using g -> gap) that have no common and require too specific context on Yoga's API that may confuse.

At first, we were following some that are already kinda normal and known in some other projects (such as mr -> margin-left). I don't know how we're gonna end if we became too permissive and end put with stuff like:

```jsx
<Box
  m="small"
  g="xlarge"
  b="small"
  bc="red" // -> background-color? border-color?
  o="1px solid red"
/>
```

What does this even means?

cc/ @alan-oliv @ggdaltoso @tcK1

tcK1:
> I don't know how we're gonna end if we became too permissive

You have a valid point! Maybe we shouldn't abbreviate this property. If in doubt, we can always look into the [Tailwind documentation](https://tailwindcss.com/docs/gap) and see how they are doing, since they already have almost (if not all) properties.

Also, what about `row` and `column` gaps? Should we also add it?

PR #375, opened 2021-10-22
# ✨ Add system support for Chips

mmartinsoliveira:
# ✨ Add system support for Chips

### Good News

News props for Chips Component:

- `backgroundColor` 
- `spacing`
- `border`
- `elevation`
- `layout`
- `typography`

### Screenshot

image

image


coelhucas:
We'll define more explicit guidelines in how to decide which properties from the system we adopt for each kind of component. Since this specifically will keep as is, I'm closing this.

PR #370, opened 2021-10-15
fix(yoga): dropdown selection cleareable

vitebo:
fix #130

PR #378, opened 2021-11-03
Fix Dropdown Issues

frgiovanna:
[jira-task]: https://gympass.atlassian.net/browse/ACCMGMT-534
[jira-img]: https://img.shields.io/badge/issue-JIRA-blue.svg

[![JIRA][jira-img]][jira-task]

## Description

The current dropdown component in Yoga has some usability issues and visual inconsistencies. In this PR we are fixing this.
[Here](https://www.figma.com/file/zjUBbl9Rfg6EhbXdReL2HN/%5BCross%5D-Chapter-design-desk?node-id=853%3A44530) you can see all the detailed specs.

## States
Default:
Screenshot from 2021-11-09 10-58-43

Open:
Screenshot from 2021-11-09 10-58-59

Selected:
Screenshot from 2021-11-09 10-59-26

Error:
Screenshot from 2021-11-09 10-59-55

Disabled:
Screenshot from 2021-11-09 11-00-08

## Changes
- Making the label visible when the dropdown is open or with a selected item
- Changing arrow, border, and text colors

### DIFF
--- a/packages/yoga/src/Dropdown/web/Dropdown.jsx
+++ b/packages/yoga/src/Dropdown/web/Dropdown.jsx
@@ -202,6 +202,7 @@ const OptionsList = styled.ul`
 const Option = styled.li`
   ${({
     isSelected,
+    highlighted,
     theme: {
       yoga: {
         baseFont,
@@ -238,6 +239,12 @@ const Option = styled.li`
         ? dropdown.selected.option.font.color
         : dropdown.option.font.color
     };
+
+    ${
+      highlighted
+        ? `background-color: ${dropdown.hover.option.backgroundColor};`
+        : ''
+    }
 
     &:hover {
       background-color: ${dropdown.hover.option.backgroundColor};
--- a/packages/yoga/src/Dropdown/web/Dropdown.jsx
+++ b/packages/yoga/src/Dropdown/web/Dropdown.jsx
@@ -113,10 +113,14 @@ const labelTransition = css`
     font-size: ${input.label.font.size.typed}px;
     color: ${dropdown.input.label.color};
 <<<<<<< HEAD
+<<<<<<< HEAD
     margin-left: ${dropdown.input.label.margin.left}px;
 =======
     margin-left: ${dropdown.input.label.margin.left}
 >>>>>>> 251d7d7cf... :sparkles: fix: change dropdown configs
+=======
+    margin-left: ${dropdown.input.label.margin.left}px;
+>>>>>>> 23df53c1c... :art: feat: add legend behavior to another scenery
   `}
 `;
 
@@ -125,9 +129,13 @@ const Input = styled.input`
     disabled,
     value,
 <<<<<<< HEAD
+<<<<<<< HEAD
     isOpen,
 =======
 >>>>>>> 251d7d7cf... :sparkles: fix: change dropdown configs
+=======
+    isOpen,
+>>>>>>> 23df53c1c... :art: feat: add legend behavior to another scenery
     theme: {
       yoga: {
         transition,
@@ -159,6 +167,9 @@ const Input = styled.input`
     }
 
 <<<<<<< HEAD
+<<<<<<< HEAD
+=======
+>>>>>>> 23df53c1c... :art: feat: add legend behavior to another scenery
     ${value || isOpen
       ? css`
           & ~ legend {
@@ -172,9 +183,12 @@ const Input = styled.input`
       : ''}
 
     ${isOpen &&
+<<<<<<< HEAD
 =======
     ${value &&
 >>>>>>> 251d7d7cf... :sparkles: fix: change dropdown configs
+=======
+>>>>>>> 23df53c1c... :art: feat: add legend behavior to another scenery
       css`
         & ~ legend {
           max-width: 1000px;
--- a/packages/yoga/src/Dropdown/web/Dropdown.jsx
+++ b/packages/yoga/src/Dropdown/web/Dropdown.jsx
@@ -101,27 +101,8 @@ const Selector = styled.div`
 `;
 
 const Field = styled(Input)`
-  ${({
-    value,
-    isOpen,
-    disabled,
-    theme: {
-      yoga: { transition },
-    },
-  }) => css`
-    &:focus {
-      & ~ legend {
-        max-width: max-content;
-        transition-property: max-width;
-        transition-duration: ${transition.duration[1]}ms;
-      }
-
-      & ~ label {
-        ${labelTransition}
-      }
-    }
-
-    ${(value || isOpen) && !disabled
+  ${({ value, isOpen, disabled }) => css`
+    ${isOpen && !disabled
       ? css`
           & ~ legend {
             max-width: max-content;
@@ -133,6 +114,7 @@ const Field = styled(Input)`
         `
       : ''}
 
+<<<<<<< HEAD
     ${isOpen &&
       css`
         & ~ legend {
@@ -143,6 +125,15 @@ const Field = styled(Input)`
           ${labelTransition};
         }
       `}
+=======
+    ${value && !disabled
+      ? css`
+          & ~ label {
+            ${labelTransition};
+          }
+        `
+      : ''}
+>>>>>>> refactor: avoiding unnecessary overwrite
   `}
 `;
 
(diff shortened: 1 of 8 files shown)


renanmpn:
@frgiovanna Can we also add some screen shot showing how was before

frgiovanna:
> @frgiovanna Can we also add some screen shot showing how was before

done!

ggdaltoso:
Hi, @frgiovanna! Thanks for your contribution!  

Just to let things more clear to me, is this PR related to #370?
As you are working in Dropdown too, could you help us review it anyway?

Also, I'm seeing you probably are using [gitmoji](https://gitmoji.dev/). Right?
At Yoga we use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/). We have a script to help us with it: 

```bash
yarn commit
``` 

You'll be prompted with some steps which will guide you to build the final commit message.

We use it to generate our changelog and versioning automatically. With that said, I would ask you if you could revisit your commits and change the messages following our [guide](https://github.com/Gympass/yoga/blob/master/CONTRIBUTING.md).

renanmpn (approved):

frgiovanna:
> Hi, @frgiovanna! Thanks for your contribution!
> 
> Just to let things more clear to me, is this PR related to #370? As you are working in Dropdown too, could you help us review it anyway?
> 
> Also, I'm seeing you probably are using [gitmoji](https://gitmoji.dev/). Right? At Yoga we use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/). We have a script to help us with it:
> 
> ```shell
> yarn commit
> ```
> 
> You'll be prompted with some steps which will guide you to build the final commit message.
> 
> We use it to generate our changelog and versioning automatically. With that said, I would ask you if you could revisit your commits and change the messages following our [guide](https://github.com/Gympass/yoga/blob/master/CONTRIBUTING.md).

Hey @ggdaltoso, 
It is related to #370, but here I'm solving more things than that PR, you can check the Jira issue [here](https://gympass.atlassian.net/browse/ACCMGMT-507). 

About the commits, my bad, I let it pass, I'll be changing this.

packages/yoga/src/Dropdown/web/Dropdown.jsx
```diff
@@ -112,11 +140,40 @@ const Input = styled.input`
     line-height: ${dropdown.input.font.lineHeight}px;
     cursor: ${disabled ? 'not-allowed' : 'pointer'};
 
-    &, &::placeholder {
-      color: ${dropdown.input.font.color};
-      ${selected ? `color: ${dropdown.selected.input.font.color};` : ''}
-      ${disabled ? `color: ${dropdown.disabled.input.font.color};` : ''}
+    &:focus {
+      & ~ legend {
+        max-width: 1000px;
```
coelhucas, line 145: These "hardcoded" `max-width`s are kinda tricky, what's the rationale behind them (so we can think on a good way to remove if possible)?

packages/yoga/src/Dropdown/web/Dropdown.jsx
```diff
@@ -325,9 +374,9 @@ const Dropdown = ({
             />
           </Button>
         </Selector>
-
         {isOpen && (
           <OptionsList selected={selectedItem !== null} {...getMenuProps()}>
+            <Option onClick={() => selectItem(null)} />
```
coelhucas, line 379: I wonder if this is the best API, since there's some cases that we need to omit the possibility of clearing the selection. As well as in some cases the user might need to see some value to the clear option, such as the default behavior for `<select>`
image


I believe we could recreate this Dropdown API (which will generate a breaking change) to achieve something as in [MUI](https://mui.com/components/selects/), for instance.

I'd like to hear everyone thoughts on this, so cc/ @ggdaltoso @tcK1. We can also bring this to further discussion including more people

packages/yoga/src/Dropdown/web/Dropdown.jsx
frgiovanna: I bring these changes from the Input component, there we have this [commit](https://github.com/Gympass/yoga/pull/164/commits/0897595f6dd6c2ef029cd741a7cfbf9b43f54180#diff-deb82c5065b3562f56a527453a2c76c3fd13479575decd49aca87caa56f3b224R59)
I'm not sure if this hardcoded max-width is necessary, I removed it and it keeps the expected behavior.

coelhucas: Good to know, I wasn't aware of this, I'll gave it a look. Thanks!

coelhucas: Just a quick update on this: I noticed that in fact a `max-width` is needed, since without it we'll have this behavior:
image

_But_, `1000px` is too hacky, what do you think about:
```suggestion
        max-width: max-content;
```

I tested with this and it works the same. You just have to update yours and I'll take care of the already existing Inputs 🌻

ggdaltoso:
I think the Dropdown shouldn't care about the empty selection. It should not be its responsibility which option the user should have.
By doing this in this way, we're forcing every dropdown to have an empty field. How would be the behavior of a required Dropdown field inside a form?

If someone who uses Yoga's Dropdown wants to have an empty field, just add it as an option and add it as a prop to `options`.

The only thing our Dropdown doesn't have, for sure, is this label behavior, as we see in the Input field as well. With this in mind, couldn't we make use of the Input element in any way? 
We take advantage of Input inside [`AutoComplete`](https://github.com/Gympass/yoga/blob/master/packages/yoga/src/AutoComplete/web/AutoComplete.jsx), just for reference.

renanmpn:
Folks. Does make sense to us to push this PR and start a parallel discussion about the API and break changes?
I understand that everyone could benefit from this improvement now. And later make the changes that would generate a breaking change.

coelhucas:
> Folks. Does make sense to us to push this PR and start a parallel discussion about the API and break changes? I understand that everyone could benefit from this improvement now. And later make the changes that would generate a breaking change.

There's stuff that we can't allow for people start using from this API, such as the "blank option" which doesn't follow any standard for de-selection models. Also, in cases where we have for instance a form full of required values, one can mistakenly send "BLANK" as a value, which isn't desired. If we need to have this merged fast, it would be good to at least remove this de-selection implementation until we have further discussion on rewriting this API (as well as the conflicts which must be solved)

cc/ @ggdaltoso WDYT?

ggdaltoso:
> Folks. Does make sense to us to push this PR and start a parallel discussion about the API and break changes? I understand that everyone could benefit from this improvement now. And later make the changes that would generate a breaking change.

This discussion should have been done earlier then opening this PR, @renanmpn, and as @coelhucas was saying, we can't force everyone to have a behavior they mustn't

coelhucas:
Your merge still isn't properly done. You're including older versions of [Navigation](https://github.com/Gympass/yoga/blob/master/packages/doc/src/components/Navigation/Navigation.jsx).

frgiovanna:
> Your merge still isn't properly done. You're including older versions of [Navigation](https://github.com/Gympass/yoga/blob/master/packages/doc/src/components/Navigation/Navigation.jsx).

Thanks for the push, I'm working to fix this. I'll open the PR and ask for your review again when it's finished

packages/yoga/src/Dropdown/web/Dropdown.jsx
```diff
@@ -306,12 +297,13 @@ const Dropdown = ({
           error={error}
           selected={selectedItem !== null}
         >
-          <Input
+          <Field
             readOnly
-            placeholder={label}
             disabled={disabled}
             selected={selectedItem !== null}
             {...getInputProps()}
```
dehmirandac2, line 304: I guess we should move this getInputProps to the last position of the Field props. This way we could overwrite isOpen or label if we need. Does it make sense?

``` js
<Field
  readOnly
  disabled={disabled}
  selected={selectedItem !== null}
  isOpen={isOpen}
  label={label}
  {...getInputProps()}
/>
```

packages/yoga/src/Dropdown/web/Dropdown.jsx
```diff
@@ -31,92 +33,95 @@ const Wrapper = styled.div`
...
+
+      & ~ label {
+        ${labelTransition}
+      }
     }
+
+    ${(value || isOpen) && !disabled
+      ? css`
+          & ~ legend {
+            max-width: max-content;
```
coelhucas, line 117: Should these styles be replicated since we're already using Yoga's Input that already contains that?

dehmirandac2 (approved):

douglasmiguel7 (approved):

renanmpn (approved):

coelhucas:
Yaaay! Welcome aboard 🎈

coelhucas:
please @allcontributors add @frgiovanna for code

PR #384, opened 2021-11-05
fix(navigation): correct ordering across browsers

coelhucas:
I messed up in #371 and I forgot to test the changes in Chrome. I assumed that it was working because it was ok on Firefox and Safari. I'm sorry for this, Chrome users 😅 

Now it is really consistent:
Firefox | Chrome | Safari
image


mmartinsoliveira (approved):
LGTM

packages/doc/src/components/Navigation/Navigation.jsx
```diff
@@ -19,9 +19,9 @@ import MDXElements from '../MDXElements';
  */
 const getSorting = kind =>
   ({
-    order: (a, b) => (a.order > b.order ? 1 : -1),
-    alphabetic: (a, b) => (a > b ? -1 : 1),
-  }[kind ?? 'order']);
+    alphabeticDescending: (a, b) => (a.title > b.title ? -1 : 1),
+    alphabeticAscending: (a, b) => (a.title > b.title ? 1 : -1),
+  }[kind ?? 'alphabeticDescending']);
```
ggdaltoso, line 24: We use `order` to sort some [documentation](https://github.com/Gympass/yoga/blob/master/packages/doc/content/guidelines/product-content/grammar-and-mechanics/index.mdx?plain=1#L5). It is also a parameter of our frontmatter schema.

How is the impact of this change on our [guidelines](https://gympass.github.io/yoga/guidelines/product-content)?

ggdaltoso (changes requested):

packages/doc/src/components/Navigation/Navigation.jsx
coelhucas: Thanks for pointing this out, it was really some lack of attention to the order defined in Markdown (We didn't noticed this in the previous PR and the guidelines page is broken since then). I've fixed it and made the code infer which sorting function to use based on the MDX having an `order` value or not. What do you think?

image

marcosricardo (approved):

ggdaltoso (approved):

PR #388, opened 2021-11-08
✨  Add Dialog

mmartinsoliveira:
# ✨ Add Dialog

> ⚠️  _This PR add support for web only_

https://user-images.githubusercontent.com/46993493/141116759-73d3bc2a-3b21-46ed-9acf-29703f3e8f5a.mov

## Motivation

The Dialog component has the objective to provide priority feedback for the user.

Considering this is high-priority feedback, this component interrupts the user journey and requires the maximum attention of the user.

This component has two props `isOpen` and `onClose`.

As it could be seen in most components, the Dialog has its own compounds:

- `Dialog.Header`
- `Dialog.Content`
- `Dialog.Actions`

## Changes

- create new component `<Dialog />`
- add the `dialog.theme`
- add snapshot

## Useful links

[Figma prototype](https://www.figma.com/file/wn9grDi9mM29Wt7LmQjol7/branch/FvsJjJxJAjgTHNAwknRNsM/Yoga-Design-System-7.1?node-id=18725%3A58521)


packages/doc/content/components/components/dialog/index.mdx
```diff
@@ -0,0 +1,168 @@
...
+title: 'Dialog'
+metaTitle: 'Dialog'
+metaDescription: 'Dialog Component'
+---
+
+# Dialog
+
+### Reference
+
+Gympass `<Dialog />` is a component has the objective o provide priority feedback for the user
```
coelhucas, line 11:
```suggestion
Gympass `<Dialog />` is a component which has the purpose o provide priority feedback for the user
```

packages/yoga/src/Dialog/web/Dialog.jsx
```diff
@@ -0,0 +1,125 @@
...
+
+  padding: ${onClose ? dialog.padding.withIconClose : dialog.padding.top}px 
+  ${dialog.padding.default}px 
+  ${dialog.padding.default}px;
+
+  width: ${dialog.width.default}px;
+  min-height: ${dialog.height.min}px;
+  border-radius: ${dialog.border.radius}px;
+
+  backdrop-filter: blur(1px);
```
coelhucas, line 31: This isn't well supported in [some browsers](https://caniuse.com/css-backdrop-filter)

packages/yoga/src/Dialog/web/Header.jsx
```diff
@@ -0,0 +1,17 @@
...
+
+const Header = props => (
+  <Card.Header
+    ta="center"
+    color="text.primary"
+    fw="medium"
+    fs="xlarge"
+    mb="large"
+    {...props}
+  />
```
coelhucas, line 12: I'm not sure if in this case it worth to use Card's Header (if we update Card's one this will break), maybe we could bring this to PD team in order to define maybe a standard for these headers? Because if it happens we can then have a separate Header which both Card and Dialog uses.

packages/yoga/src/Card/web/Card/Content.jsx
```diff
@@ -1,6 +1,7 @@
 import styled from 'styled-components';
+import Box from '../../../Box';
 
-const Content = styled.div``;
+const Content = styled(Box)``;
```
tcK1, line 4:
```suggestion
const Content = Box;
```

packages/yoga/src/Dialog/web/Dialog.jsx
```diff
@@ -0,0 +1,123 @@
...
+            </Box>
+          )}
+          {children}
+        </StyledDialog>
+      </Overlay>,
+      dialogElement,
+    )
+  ) : (
+    <> </>
+  );
```
tcK1, line 102: I think using `null` is a better approach than using a Fragment. Although it is empty, it is another node on the React Tree.

```suggestion
  ) : null;
```

packages/yoga/src/Dialog/web/index.js
```diff
@@ -0,0 +1,10 @@
+import Dialog from './Dialog';
+import Header from './Header';
+import Content from './Content';
+import Actions from './Actions';
+
+Dialog.Header = Header;
+Dialog.Content = Content;
+Dialog.Actions = Actions;
```
tcK1, line 8: I think that instead of Actions, we could call it Footer, to be more generic. Is this a bad idea?

packages/yoga/src/hooks/index.js
```diff
@@ -0,0 +1,3 @@
+import usePortal from './usePortal';
```
tcK1, line 1: If we want to make this `index` file easier to maintain, maybe we can change some things.

What about exporting the hook from the `usePortal.jsx` file as a named export, and just re-export it here, like this:

```js
// At hooks/usePortal.jsx

export const usePortal = () => { /* ... */ };

// At hooks/index.js

export * from './usePortal';
```

This way we only need to add 1 line to this file, and if we use something like [`glob`](https://www.npmjs.com/package/import-glob) in the future, it would be even easier. 😊 

What do you think?

packages/yoga/src/Dialog/web/index.js
mmartinsoliveira: I think it's a great idea, but I followed the same model as the card's components.

[Example](https://gympass.github.io/yoga/components/card/default)


But I can follow up with your suggestion, which I liked.

packages/yoga/src/Dialog/web/Dialog.jsx
```diff
@@ -0,0 +1,123 @@
...
+
+const Dialog = ({ isOpen, children, onClose }) => {
+  const dialogRef = useRef(null);
+
+  const dialogElement = usePortal('dialog');
+
+  const closeDialog = useCallback(
+    e => {
+      if (dialogRef.current === e.target && isOpen) {
+        onClose(false);
```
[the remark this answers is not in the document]
mmartinsoliveira: it makes perfect sense 🚀

packages/yoga/src/hooks/index.js
mmartinsoliveira: Great idea 👌

tcK1 (approved):

mavericardo (approved):

coelhucas (approved):

tgomonteiro (approved):

coelhucas:
🚀

PR #402, opened 2021-11-12
Update main job and remove bloat from `npmpublish` workflow

coelhucas:
- Use the correct lint command in `linter_and_tests` job
- Remove redundant steps from `npmpublish` (every PR will already have them, so we don't need to repeat it) job in order to make it faster


tcK1 (approved):

frgiovanna (approved):
:rocket:

PR #407, opened 2021-11-17
Change the limitLabel defaultProps and remove the 0 from numberOfItemsLeft property

alycecristines:
## Description 📄

When using this component without passing anything in the limitLabel property the text undefined was
showing and this PR fixes this problem. Also removed the padding 0 from the numberOfItemsLeft property. 

## Screenshots 📸
| Before |  After | 
| ------ | ------ | 
|    image    |        image    |

### DIFF
--- a/packages/yoga/src/Result/native/Details.jsx
+++ b/packages/yoga/src/Result/native/Details.jsx
@@ -83,7 +83,7 @@ const ResultDetails = ({
                   <>
                     <Separator />
                     <StyledText variant={variant}>
-                      + {numberOfItemsLeft.toString().padStart(2, '0')}
+                      + {numberOfItemsLeft.toString()}
                       {limitLabel !== '' && ` ${limitLabel}`}
                     </StyledText>
                     <Separator />
@@ -117,7 +117,7 @@ ResultDetails.propTypes = {
 
 ResultDetails.defaultProps = {
   limit: undefined,
-  limitLabel: undefined,
+  limitLabel: '',
   dots: false,
   renderItem: Text.Tiny,
 };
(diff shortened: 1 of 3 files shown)


ericcleao (approved):

packages/yoga/src/Result/native/Details.jsx
```diff
@@ -83,7 +83,7 @@ const ResultDetails = ({
                   <>
                     <Separator />
                     <StyledText variant={variant}>
-                      + {numberOfItemsLeft.toString().padStart(2, '0')}
+                      + {numberOfItemsLeft.toString()}
                       {limitLabel !== '' && ` ${limitLabel}`}
```
ericcleao, line 87: Why do not change to:
```jsx
{!!limitLabel ? '' : `{limitLabel}`}
```
Or just

```jsx
{!!limitLabel && `{limitLabel}`}
```
Once we should not show nether `null`, empty or `undefined` as limit label.

ericcleao: Thinking in cases that the `limitLabel` prop is set as `null` or `undefined` for instance (not just the default prop value).

packages/yoga/src/Result/native/Details.jsx
```diff
@@ -83,7 +83,7 @@ const ResultDetails = ({
                   <>
                     <Separator />
                     <StyledText variant={variant}>
-                      + {numberOfItemsLeft.toString().padStart(2, '0')}
+                      + {numberOfItemsLeft.toString()}
```
leonardokl, line 86:
```suggestion
                      + {numberOfItemsLeft}
```

leonardokl (approved):

coelhucas:
I think this component could get snapshots for other "variants" of its props, for instance: this change should break a snapshot in order to let us know that it had a visual regression (removing the leading zeros) so we could re-generate the snapshot with confidence that it would catch visual issues.

katharinedepaulo (approved):

PR #410, opened 2021-11-25
feat(component): changed the design of App PlanCard to new layout

salomao-neto:
Changed the PlanCard design to new layout with the Price and Title in same size and with the possibility of change the ListItem color.
[JIRA](https://gympass.atlassian.net/browse/ENJ-388)
[ FIGMA](https://www.figma.com/file/WI7iB8gTsIrY4WhWybpDwN/%5BHercules%5D-Product-Restrictions?node-id=3408%3A113709)

| Before | After |
| ------| -------| 
|Captura de Tela 2021-11-25 às 09 36 27 |Captura de Tela 2021-11-25 às 09 38 37

### DIFF
--- a/packages/doc/content/components/components/card/plancard-native.mdx
+++ b/packages/doc/content/components/components/card/plancard-native.mdx
@@ -35,12 +35,12 @@
 
 #### Props - PlanCard
 
-<PropsTable component="PlanCard" platform="web" />
+<PropsTable component="PlanCard" platform="native" />
 
 #### Props - PlanCard.Content
 
-<PropsTable component="PlanCard.Content" platform="web" />
+<PropsTable component="PlanCard.Content" platform="native" />
 
 #### Props - PlanCard.ListItem
 
-<PropsTable component="PlanCard.ListItem" platform="web" />
+<PropsTable component="PlanCard.ListItem" platform="native" />
--- a/packages/doc/content/components/components/card/plancard-native.mdx
+++ b/packages/doc/content/components/components/card/plancard-native.mdx
@@ -11,7 +11,7 @@
   >
     <PlanCard.Subtitle>Get access to</PlanCard.Subtitle>
     <PlanCard.List>
-      <PlanCard.ListItem color="#6B6B78" icon={MapPin} text="2.900 gyms and studios" />
+      <PlanCard.ListItem variant="light" icon={MapPin} text="2.900 gyms and studios" />
       <PlanCard.ListItem icon={Smartphone} text="24 wellness app" />
       <PlanCard.ListItem
         icon={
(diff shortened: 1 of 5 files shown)


packages/labnative/pages/Card.jsx
```diff
@@ -107,7 +107,11 @@ const CardPage = () => {
           >
             <PlanCard.Subtitle>Get access to</PlanCard.Subtitle>
             <PlanCard.List>
-              <PlanCard.ListItem icon={MapPin} text="2.900 gyms and studios" />
+              <PlanCard.ListItem
+                color="#6B6B78"
```
ericcleao, line 111: It's better to follow the preference order below, from left to right:

| theme color | token | hex |
| ------------ | ------ | --- |
| `text.secondary` | `deep` | `#6B6B78 ` |

packages/doc/content/components/components/card/plancard-native.mdx
```diff
@@ -11,7 +11,7 @@
   >
     <PlanCard.Subtitle>Get access to</PlanCard.Subtitle>
     <PlanCard.List>
-      <PlanCard.ListItem icon={MapPin} text="2.900 gyms and studios" />
+      <PlanCard.ListItem color="#6B6B78" icon={MapPin} text="2.900 gyms and studios" />
```
ericcleao, line 14: It's better to follow the preference order below, from left to right:

| theme color | token | hex |
| ------------ | ------ | --- |
| `text.secondary` | `deep` | `#6B6B78 ` |

ericcleao (commented):
Just some doc adjustments! 🚀

ericcleao (changes requested):
It is required that the Yoga 7.1 Figma document stay aligned with the implemented code before merging this PR.

packages/yoga/src/Card/native/PlanCard/List.jsx
```diff
@@ -72,11 +87,13 @@ ListItem.propTypes = {
   /** if provided displays a button below the item text. It accepts all button
    * element props */
   buttonProps: shape({}),
+  variant: string,
```
coelhucas, line 90: I believe it would be nice to document this property (by commenting a line before like other examples) as also adding a prop validation to check if it is included in our color tokens, what do you think?

salomao-neto: Nice, I set a little commentary above the props, about the validation, in the line 51 it will try to get the color from tokens, if it not found any color, the variable `itemColor` will be `undefined`, so in styled component `ItemText` will set a default color.

coelhucas: Makes sense, but what do you think about throwing a warning in this case so the developer knows that they're giving an invalid value?

E.g.:
```jsx
<Comp variant="deprecated-token" />
```

Since `variant` isn't undefined but the token doesn't exist, a warning such as 'Invalid token "deprecated-token", defaulting to "default-value-name"' could help one debugging what they did wrong and why their token is being overridden.

coelhucas (approved):

ericcleao (approved):

PR #411, opened 2021-11-25
feat(component): a Divider component to separate two parts of another component

salomao-neto:
The Divider component can be used to divide two parts of another component or to separate parts of
screen.

| WEB | APP |
| ----- | ----- |
|Captura de Tela 2021-11-25 às 10 14 35|Simulator Screen Shot - iPhone 11 - 2021-11-25 at 09 51 04

### DIFF
--- a/packages/doc/content/components/components/divider/index.mdx
+++ b/packages/doc/content/components/components/divider/index.mdx
@@ -1,0 +1,25 @@
+---
+title: 'Divider'
+metaTitle: 'Divider'
+metaDescription: 'Divider Component'
+---
+
+
+import DividerWeb from './divider-web.mdx';
+import DividerNative from './divider-native.mdx';
+
+
+# Divider
+
+
+<TabbedView>
+  <Tab title="Web">
+    <DividerWeb />
+  </Tab>
+
+  <Tab title="Native">
+    <DividerNative />
+  </Tab>
+</TabbedView>
+
+  
--- a/packages/yoga/src/Divider/Divider.theme.js
+++ b/packages/yoga/src/Divider/Divider.theme.js
@@ -1,0 +1,3 @@
+const Divider = () => ({});
+
+export default Divider;
--- a/packages/yoga/src/Divider/native/Divider.jsx
+++ b/packages/yoga/src/Divider/native/Divider.jsx
@@ -1,33 +1,37 @@
+import { bool } from 'prop-types';
 import React from 'react';
-import styled, { css } from 'styled-components';
-import theme from '../../Theme/helpers/themeReader';
+import styled, { css, withTheme } from 'styled-components';
+// ${vertical ? 'height: 100%' : 'width: 100%'};
 
 const StyledDivider = styled.View`
-  ${props => {
-    const { borders, colors } = theme(props);
-
+  ${({
+    theme: {
+      yoga: { borders, colors },
+    },
+    vertical,
+  }) => {
     return css`
-      height: ${borders.small};
-      width: 100%;
-      background-color: ${colors.light};
+      ${vertical ? 'height: auto' : 'width: auto'};
+      ${vertical
+        ? `border-left-width: ${borders.small}`
+        : `border-bottom-width: ${borders.small}`};
+
+      border-color: ${colors.text.disabled};
     `;
   }}
 `;
 
-const DividerContainer = styled.View`
-  flex-direction: row;
-`;
-
 const Divider = props => {
-  return (
-    <DividerContainer>
-      <StyledDivider {...props} />
-    </DividerContainer>
-  );
+  return <StyledDivider {...props} />;
 };
 
-Divider.propTypes = {};
+Divider.propTypes = {
+  /** If this value is defined, the divider will be in vertical when the flexDirection is row type */
+  vertical: bool,
+};
 
-Divider.defaultProps = {};
+Divider.defaultProps = {
+  vertical: false,
+};
 
-export default Divider;
+export default withTheme(Divider);
--- a/packages/yoga/src/Divider/native/Divider.jsx
+++ b/packages/yoga/src/Divider/native/Divider.jsx
@@ -1,0 +1,33 @@
+import React from 'react';
+import styled, { css } from 'styled-components';
+import theme from '../../Theme/helpers/themeReader';
+
+const StyledDivider = styled.View`
+  ${props => {
+    const { borders, colors } = theme(props);
+
+    return css`
+      height: ${borders.small};
+      width: 100%;
+      background-color: ${colors.light};
+    `;
+  }}
+`;
+
+const DividerContainer = styled.View`
+  flex-direction: row;
+`;
+
+const Divider = props => {
+  return (
+    <DividerContainer>
+      <StyledDivider {...props} />
+    </DividerContainer>
+  );
+};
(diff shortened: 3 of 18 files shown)


packages/yoga/src/Divider/Divider.theme.js
```diff
@@ -0,0 +1,3 @@
+const Divider = () => ({});
```
ericcleao, line 1: I notice that there are some components that do not have a theme defined but we still let the file empty, we really need it, @coelhucas?

packages/yoga/src/Divider/native/Divider.jsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import styled, { css } from 'styled-components';
+import theme from '../../Theme/helpers/themeReader';
```
ericcleao, line 3: I don't know if it is a pattern but we could avoid getting the theme from `themeReader ` helper by just updating the props from line 6.

packages/yoga/src/Divider/native/Divider.jsx
```diff
@@ -0,0 +1,33 @@
...
+
+const StyledDivider = styled.View`
+  ${props => {
+    const { borders, colors } = theme(props);
```
ericcleao, line 7:
```suggestion
  ${({
    theme: {
      yoga: {
        borders,
        colors,
      }
    },
  }) => {
```

packages/yoga/src/Divider/native/Divider.jsx
```diff
@@ -0,0 +1,33 @@
...
+
+const Divider = props => {
+  return (
+    <DividerContainer>
+      <StyledDivider {...props} />
+    </DividerContainer>
+  );
+};
+
+Divider.propTypes = {};
```
ericcleao, line 29: It should not have really any prop defined for this component?

packages/yoga/src/Divider/native/Divider.jsx
```diff
@@ -0,0 +1,33 @@
...
+`;
+
+const DividerContainer = styled.View`
+  flex-direction: row;
+`;
+
+const Divider = props => {
+  return (
+    <DividerContainer>
+      <StyledDivider {...props} />
```
ericcleao, line 24: The fact that we are passing the `props` forward made me remember about the `yoga-system`, we do not need any `system` implementation at the `Divider`?

packages/yoga/src/Divider/native/Divider.jsx
```diff
@@ -0,0 +1,33 @@
...
+import styled, { css } from 'styled-components';
+import theme from '../../Theme/helpers/themeReader';
+
+const StyledDivider = styled.View`
+  ${props => {
+    const { borders, colors } = theme(props);
+
+    return css`
+      height: ${borders.small};
+      width: 100%;
```
ericcleao, line 11: If we need to add margins to this component the `width: 100%` prop will move the component but it will stay with the same size, causing a strange behaviour, in the past, we needed to add to other components the prop `flex: 1` instead of putting the `width: 100%` to avoid this behaviour.
Is it the same case?

packages/yoga/src/Divider/native/Divider.jsx
```diff
@@ -0,0 +1,33 @@
...
+      background-color: ${colors.light};
+    `;
+  }}
+`;
+
+const DividerContainer = styled.View`
+  flex-direction: row;
```
ericcleao, line 18: Do we need to set the `display` to `flex`? There are any possible `system` props that could be used here?

packages/yoga/src/Divider/web/Divider.jsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import styled, { css } from 'styled-components';
+import theme from '../../Theme/helpers/themeReader';
```
ericcleao, line 3: Same comment from the `native` implementation.

packages/yoga/src/Divider/web/Divider.jsx
```diff
@@ -0,0 +1,25 @@
...
+
+const StyledDivider = styled.div`
+  ${props => {
+    const { borders, colors } = theme(props);
+
+    return css`
+      flex: 1;
+      height: ${borders.small}px;
+      background-color: ${colors.light};
```
ericcleao, line 12: Could we use this same approach at the `native`? Avoiding have a container there

packages/yoga/src/Divider/web/Divider.jsx
```diff
@@ -0,0 +1,25 @@
...
+  }}
+`;
+
+const Divider = props => {
+  return <StyledDivider {...props} />;
+};
+
+Divider.propTypes = {};
+
+Divider.defaultProps = {};
```
ericcleao, line 23: Same comments about the need for some props and implementing some `system` props.

packages/yoga/src/Divider/native/Divider.jsx
salomao-neto: Actually, the `width: 100%` has a size based on the father component, so, if we put a margin on father component, it will be with 100% of the size inside the margin, like the following image. (I put more height to divider just  to a better visualization)
https://reactnative.dev/docs/height-and-width#percentage-dimensions
image

packages/yoga/src/Divider/web/Divider.jsx
salomao-neto: I tried to do the same approach on `native`, but unfortunately, it not had work. Have some idea?

packages/yoga/src/Divider/Divider.theme.js
coelhucas: For some reason my comments were lost 🤔 

---
This is something generated by our CLI, it indeed isn't needed in this and some other cases, but doesn't give us any real impact in the final bundling. Up to the developer keeping or not (although it feels "useless", a component missing this file could confuse a new collaborator who doesn't know how theming works here)

We're open to define a standard to this if you want :)

coelhucas:
I'll need some time before reviewing it if it's not a problem for your team, I'll still discuss about it with the PD team. I'm not sure if this is truly needed since we can achieve this using Box (with one of its purposes is to separate components)

image

Just let you know in case I just make a further review/approval next week.

packages/doc/content/components/components/divider/index.mdx
```diff
@@ -0,0 +1,25 @@
...
+  <Tab title="Web">
+    <DividerWeb />
+  </Tab>
+
+  <Tab title="Native">
+    <DividerNative />
+  </Tab>
+</TabbedView>
+
+
```
coelhucas, line 25: I believe that this documentation in general is too weak right now. As an effort to ensure engineers can rely more on documentation we're bringing new standards for this.

While it's not explicit defined, we can use references such as [Spectrum's divider](https://opensource.adobe.com/spectrum-web-components/components/divider/). It is a simple component but I think this may make the difference to someone contributing either from inside or outside Gympass, what are your thoughts on this?

coelhucas:
This component is missing a `vertical` variation as specified by our designers. Could we include it in order to add this component to the design system without new backlog chores?

coelhucas (approved):

ericcleao (approved):

PR #419, opened 2021-12-16
fix(dropdown): fix behavior when we have the error prop

frgiovanna:
Was noticed that the last version of the dropdown component has a weird behavior when we pass an error text to it. 
This PR fixes it. 

**Changes:**
- Fix arrow Icon behavior
- Change label color 

**Screenshots**

Before:
Screenshot from 2021-12-16 14-49-34

After:
Screenshot from 2021-12-16 14-47-29

### DIFF
--- a/packages/yoga/src/Dropdown/Dropdown.theme.js
+++ b/packages/yoga/src/Dropdown/Dropdown.theme.js
@@ -26,7 +26,6 @@ const Dropdown = ({
   },
   arrow: {
     fill: colors.text.secondary,
-    error: colors.verve,
   },
   input: {
     font: {
--- a/packages/yoga/src/Dropdown/Dropdown.theme.js
+++ b/packages/yoga/src/Dropdown/Dropdown.theme.js
@@ -26,6 +26,7 @@ const Dropdown = ({
   },
   arrow: {
     fill: colors.text.secondary,
+    error: colors.verve,
   },
   input: {
     font: {
@@ -129,6 +130,7 @@ const Dropdown = ({
     selector: {
       border: {
         color: colors.text.primary,
+        error: colors.verve,
       },
     },
     option: {
--- a/packages/yoga/src/Dropdown/web/Dropdown.jsx
+++ b/packages/yoga/src/Dropdown/web/Dropdown.jsx
@@ -251,7 +251,6 @@ const ArrowIcon = styled(({ isOpen, selected, ...props }) => (
     isOpen,
     selected,
     disabled,
-    error,
     theme: {
       yoga: {
         components: { dropdown },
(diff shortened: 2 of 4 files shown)


packages/yoga/src/Dropdown/web/Dropdown.jsx
```diff
@@ -76,14 +76,28 @@ const Selector = styled.div`
...
+                border-color: ${dropdown.hover.selector.border.error};
+              }
+            }`
+              : ` 
+              &:hover {
+              fieldset {
+                border-color: ${dropdown.hover.selector.border.color};
+              }
+            }`
           }
```
tcK1, line 101: Maybe we could move the conditional to where the change happens. Would this work?

```suggestion
          &:hover {
            fieldset {
              border-color: ${
                error && !isOpen
                  ? dropdown.hover.selector.border.error
                  : dropdown.hover.selector.border.color
              };
            }
          }
```

frgiovanna: It works, and seem so much clear!

alan-oliv (changes requested):
@frgiovanna I noticed that you changed the color of the label and the arrow inside the dropdown in case of error, when according to the original component only the helper is orange as we can see [here](https://www.figma.com/file/wn9grDi9mM29Wt7LmQjol7/Yoga-Design-System-7.1?node-id=4623%3A2657)

we cannot change the original styles of the component in the implementation when the specification on figma says otherwise, please, can you review with your PD along with yoga pd's?

frgiovanna:
> @frgiovanna I noticed that you changed the color of the label and the arrow inside the dropdown in case of error, when according to the original component only the helper is orange as we can see [here](https://www.figma.com/file/wn9grDi9mM29Wt7LmQjol7/Yoga-Design-System-7.1?node-id=4623%3A2657)
> 
> we cannot change the original styles of the component in the implementation when the specification on figma says otherwise, please, can you review with your PD along with yoga pd's?

Sorry, I noticed that I was following a non-official spec. I'm changing back and keeping only the arrow position fix

mmartinsolliv:
Hi @frgiovanna! Thanks for your contribution! 

How about using conventional commits? We have a script that helps us with this

```bash
yarn commit
```

frgiovanna:
> Hi @frgiovanna! Thanks for your contribution!
> 
> How about using conventional commits? We have a script that helps us with this
> 
> ```shell
> yarn commit
> ```

Hey, I'm already using the `yarn commit`

alan-oliv (approved):

packages/yoga/src/Dropdown/Dropdown.theme.js
```diff
@@ -129,6 +129,7 @@ const Dropdown = ({
     selector: {
       border: {
         color: colors.text.primary,
+        error: colors.verve,
```
mmartinsolliv, line 132: There is this way to use it too, if you think better
 
 ```bash
 error: colors.feedback.attention.dark
 ```

mmartinsolliv (approved):

PR #423, opened 2022-01-13
refactor(input): change color of label and border

frgiovanna:
# Description

In order to follow the last design of the Input and keep the consistency of our fields, this PR changes the **color of the label in the error state** and the **color of the border in the activated state**.

You can check the design specs [here](https://www.figma.com/file/wn9grDi9mM29Wt7LmQjol7/Yoga-Design-System-7.1?node-id=4670%3A4105).

# Screenshots 

Before: 
Screenshot from 2022-01-13 17-25-14Screenshot from 2022-01-13 17-25-27

After:
Screenshot from 2022-01-13 17-32-17Screenshot from 2022-01-13 17-24-15

### DIFF
--- a/packages/yoga/src/Input/web/Field.jsx
+++ b/packages/yoga/src/Input/web/Field.jsx
@@ -74,8 +74,11 @@ const Field = styled.input`
 
         font-weight: ${input.label.font.weight};
         color: ${
-          error ? `${colors.feedback.attention[1]}` : `${colors.text.primary}`
+          error || (value && error)
+            ? `${colors.feedback.attention[1]}`
+            : `${colors.text.primary}`
         };
+
       }
     }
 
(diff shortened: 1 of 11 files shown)


Wendlereis (approved):

packages/yoga/src/Input/web/Field.jsx
```diff
@@ -74,8 +74,11 @@ const Field = styled.input`
 
         font-weight: ${input.label.font.weight};
         color: ${
-          error ? `${colors.feedback.attention[1]}` : `${colors.text.primary}`
+          error || (value && error)
```
tcK1, line 77: I think that this expression can be simplified to only: `error`. Like it was before.

1. If `error` value is **truthy**, the expression is **truthy** on the first condition.
2. if `error` is value is **falsy**, both `error` and `value && error` will be **falsy**

tcK1 (changes requested):

packages/yoga/src/Input/web/Field.jsx
frgiovanna: Thanks for the push, I'm moving this condition to the right place.

tcK1 (approved):

mmartinsoliv (approved):

PR #438, opened 2022-02-11
✨ Clipboard

Falkaniere:
https://user-images.githubusercontent.com/39073602/153646711-2feabaf1-c1ab-471f-930d-dae359c0097b.mov

## Motivation

- Facility to copy some text to test

## Changes

- create new component to copy text

### DIFF
--- a/packages/doc/src/components/Clipboard/index.jsx
+++ b/packages/doc/src/components/Clipboard/index.jsx
@@ -14,17 +14,12 @@ const Clipboard = ({ copyText }) => {
     return textToCopy;
   }
 
-  const handleCopyClick = () => {
-    CopyTextToClipboard(copyText)
-      .then(() => {
-        setIsCopied(true);
-        setTimeout(() => {
-          setIsCopied(false);
-        }, 1500);
-      })
-      .catch(error => {
-        throw new Error(error);
-      });
+  const handleCopyClick = async () => {
+    await copyTextToClipboard(copyText);
+    setIsCopied(true);
+    setTimeout(() => {
+      setIsCopied(false);
+    }, 1500);
   };
 
   return (
--- a/packages/doc/src/components/Clipboard/index.jsx
+++ b/packages/doc/src/components/Clipboard/index.jsx
@@ -8,7 +8,7 @@ import { CheckedFull, Copy } from '@gympass/yoga-icons/src';
 const Clipboard = ({ copyText }) => {
   const [isCopied, setIsCopied] = useState(false);
 
-  async function CopyTextToClipboard(text) {
+  async function copyTextToClipboard(text) {
     const textToCopy = await navigator.clipboard.writeText(text);
 
     return textToCopy;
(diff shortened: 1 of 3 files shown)


mmartinsoliv (approved):
LGTM 🚀

ericcleao (approved):

packages/doc/src/components/Clipboard/index.jsx
```diff
@@ -0,0 +1,49 @@
...
+import React, { useState } from 'react';
+import { Button, Box } from '@gympass/yoga';
+import { string } from 'prop-types';
+
+import { CheckedFull, Copy } from '@gympass/yoga-icons/src';
+
+const Clipboard = ({ copyText }) => {
+  const [isCopied, setIsCopied] = useState(false);
+
+  async function CopyTextToClipboard(text) {
```
tcK1, line 11:
```suggestion
  async function copyTextToClipboard(text) {
```

packages/doc/src/components/Clipboard/index.jsx
```diff
@@ -0,0 +1,49 @@
...
+    CopyTextToClipboard(copyText)
+      .then(() => {
+        setIsCopied(true);
+        setTimeout(() => {
+          setIsCopied(false);
+        }, 1500);
+      })
+      .catch(error => {
+        throw new Error(error);
+      });
```
tcK1, line 27: Maybe we could write this using `async`/`await`. I also think that we don't need the catch block if we are just throwing the error again.

```suggestion
    await copyTextToClipboard(copyText);
    setIsCopied(true);
    setTimeout(() => {
      setIsCopied(false);
    }, 1500);
```

mavericardo (approved):

packages/doc/src/components/Clipboard/index.jsx
Falkaniere: make sense! Thanks!

packages/doc/src/components/Clipboard/index.jsx
```diff
@@ -0,0 +1,44 @@
...
+
+    return textToCopy;
+  }
+
+  const handleCopyClick = async () => {
+    await copyTextToClipboard(copyText);
+    setIsCopied(true);
+    setTimeout(() => {
+      setIsCopied(false);
+    }, 1500);
```
Wendlereis, line 22: There is a very known good practice when using setTimeout. 

- You need to clean the timeout every time the component is unmounted to avoid side effects.
- You may need to store the timeout on a `React.ref` to be able to clean it afterward

You can check more details on this [article](https://felixgerschau.com/react-hooks-settimeout/). Also, you can check a Yoga usage [here](https://github.com/gympass/yoga/blob/master/packages/yoga/src/Snackbar/web/Snackbar.jsx#L115-L125)

Wendlereis (changes requested):

Wendlereis (approved):
Perfect!

PR #440, opened 2022-02-14
🎨  Adjust snackbar

Wendlereis:
# 🎨  Adjust snackbar

## Motivation
We noticed that our current `<Snackbar/>` was missing a behavior. We would like to show the close button even when we've added the `duration` prop to make it close automatically

## Changes
- add `hideCloseButton` prop to hide the button when the user needs
- add tests
- adjust snapshot
- adjust documentation

## Useful links
[Figma prototype](https://www.figma.com/file/wn9grDi9mM29Wt7LmQjol7/Yoga-Design-System-7.1?node-id=15654%3A44275)

### DIFF
--- a/packages/yoga/src/Snackbar/web/Snackbar.jsx
+++ b/packages/yoga/src/Snackbar/web/Snackbar.jsx
@@ -103,6 +103,7 @@ const Snackbar = ({
   actionLabel,
   onAction,
   onClose,
+  hideCloseButton,
   theme: {
     yoga: {
       components: { snackbar },
@@ -154,9 +155,9 @@ const Snackbar = ({
             </Button.Link>
           )}
 
-          {!duration && onClose && (
+          {!hideCloseButton && onClose && (
             <IconButtonWrapper role="button" onClick={onClose}>
-              <Icon as={Close} fill="secondary" width="large" height="large" />
+              <Icon as={Close} fill="secondary" size="medium" />
             </IconButtonWrapper>
           )}
         </ActionsWrapper>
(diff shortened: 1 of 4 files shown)


packages/yoga/src/Snackbar/web/Snackbar.jsx
```diff
@@ -154,9 +155,9 @@ const Snackbar = ({
             </Button.Link>
           )}
 
-          {!duration && onClose && (
+          {!hideCloseButton && onClose && (
```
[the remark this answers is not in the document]
dehmirandac2: sounds good to me!

dehmirandac2 (approved):

packages/yoga/src/Snackbar/web/Snackbar.test.jsx
```diff
@@ -102,4 +109,25 @@ describe('<Snackbar />', () => {
 
     expect(onCloseMock).toHaveBeenCalledTimes(1);
   });
+
+  it('should close with timeout set up', async () => {
```
LeoSilvaGomes, line 113: should we start to use 'test' instead 'it' here too?

LeoSilvaGomes (approved):
🎉

packages/yoga/src/Snackbar/web/Snackbar.test.jsx
Wendlereis: For this project, `it` is the standard

mmartinsoliv (approved):

Issue #418, opened 2021-12-10
[FEAT] Link component render as anchor html tag


dehmirandac2:
**Is your feature request related to a problem? Please describe.**
The Link component renders as a button HTML tag so we don't have an option to use it as an anchor. When we want to use the Link component to work as a link (open an external URL, for example) we have to implement a logic like this:

``` js
const handleClick = () => {
   window.location.replace('http://www.google.com');
}

<Button.Link onClick={handleClick}>Open external url</Button.Link>
```

**Describe the solution you'd like**
The idea is that we have the Link component rendering as an anchor and its attributes (href, target, etc). This way we can use the component as closes as a link component should be. Example:

``` js
<Button.Link href="http://www.google.com" target="blank">Open external url</Button.Link>
``` 

**Describe alternatives you've considered**

**First alternative**

An option to add this functionality without adding a breaking change is adding rendering the button as an anchor only if the href props is present. Something like this:

``` js
// How to use the tagName prop in the Button.Link
const Button = ({ href, ...otherProps }) => {
  if (href) {
     return <a href={href} {...otheProps}  />;
  }
  return <button {...otherProps} />;
};


// How to use it
<Button.Link href="http://www.google.com" target="blank">Open external url</Button.Link>
``` 

> Reference [using styled-components "as"](https://styled-components.com/docs/api#as-polymorphic-prop)

This way we can keep the default behavior as a button HTML tag. You can check the [Material UI do something like this](https://mui.com/components/buttons/#text-button)

**Second alternative**

Create a new component to the Link as an anchor and do not change the Button.Link component

``` js
// new component
<Link href="http://www.google.com" target="blank">Open external url</Link>

// old component (not changed) 
<Button.Link onClick={() => console.log('click')}>button</Button.Link>

``` 

**Additional context**
You can check how Material UI creates the [Link component here](https://mui.com/components/links/)

ricardomalan:
Hi @dehmirandac2, I was looking at the project and I believe there was a mistake in creating the `Button.Link`, it was created as `styled.button` and it would actually be `styled.a`, because then it would receive the props of a link, like: **href**, **target**, **rel** ...

| StyledButton Before - packages > yoga > src > Button > web > StyledButton.jsx  |
| --- |
![image](https://user-images.githubusercontent.com/53584223/151561573-cef77b58-7cde-4876-8284-e39565e7a605.png)

| StyledButton After - packages > yoga > src > Button > web > StyledButton.jsx |
| --- |
[...]

PR #447, opened 2022-03-10
feat(button link): add href prop to the link button render as an anchor

dehmirandac2:
## Motivation
To fix the [issue 418](https://github.com/Gympass/yoga/issues/418) I'm adding a new prop to the `Button.Link` component.
If you add the `href` prop to the component it will render as an anchor.

Example:
```js
<Button.Link href="http://www.google.com" target="blank">Open external url</Button.Link>
```

How the component will be rendered:
```js
<a href="http://www.google.com" target="blank">Open external url</a>
```

### Default value is an anchor
Screen Shot 2022-04-06 at 09 36 25

### To render as a button, you should add the "as"
Screen Shot 2022-04-06 at 09 36 32

### Table of props
Screen Shot 2022-04-06 at 09 36 39

## Changes
- Using [styled component "as" ](https://styled-components.com/docs/api#as-polymorphic-prop)to render the the component as an anchor element by default
- If the user want to use as button, they should add `as="button"`
- Add snapshot test
- Add more info to the documentation

### DIFF
--- a/packages/doc/content/components/components/button/link-web.mdx
+++ b/packages/doc/content/components/components/button/link-web.mdx
@@ -1,18 +1,20 @@
-### Usage default
-
-```javascript
-<Box display="flex" justifyContent="space-evenly" width="100%">
-  <Button.Link>Find an activity</Button.Link>
-  <Button.Link secondary>Find an activity</Button.Link>
-</Box>
-```
-
-#### Using as an anchor
+### Usage default (as an anchor)
 
 ```javascript
 <Box display="flex" justifyContent="space-evenly" width="100%">
   <Button.Link href="http://www.gympass.com" target="blank">
     Find an activity (external link)
+  </Button.Link>
+</Box>
+```
+
+#### Using as a button
+
+```javascript
+<Box display="flex" justifyContent="space-evenly" width="100%">
+  <Button.Link as="button">Find an activity</Button.Link>
+  <Button.Link as="button" secondary>
+    Find an activity
   </Button.Link>
 </Box>
 ```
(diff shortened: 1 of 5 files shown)


ericcleao (approved):

packages/doc/content/components/components/button/link-web.mdx
```diff
@@ -7,6 +7,16 @@
 </Box>
 ```
 
+#### Using as an anchor
```
Wendlereis, line 10: Would be nice to push this as recommended usage for this component. How could we create a deprecation strategy here?

cc: @mmartinsoliv @tcK1 @dehmirandac2

frgiovanna (approved):

Wendlereis:
I have noticed that we have 2 issues for this PR. One of them has a suggestion to make any button behaves as a `<a/>` given a `href` prop. 

Do you think worth add this on this PR?

dehmirandac2:
> I have noticed that we have 2 issues for this PR. One of them has a suggestion to make any button behaves as a `<a/>` given a `href` prop.
> 
> Do you think worth add this on this PR?

@Wendlereis I'm extending this functionality to all Buttons [in this PR](https://github.com/Gympass/yoga/pull/449). If you could check, please

mmartinsoliv:
Hi, @dehmirandac2 thank you so much for your contribution! 

I'm thinking about this PR here and what do you think about us keeping the `Button.Link` with the `a` tag as the default?

For the other Buttons, we can keep the `button` tag as default, if we need the default `Button` to be with the `a` tag, we can use the polymorphic prop.

`styled-components` API example: [doc](https://styled-components.com/docs/api#as-polymorphic-prop).

What do you think?

PR #449, opened 2022-03-17
Feat/buttons as anchor

dehmirandac2:
## Motivation
To extend the task #447 I'm adding a new prop to the `Button` component so it can work as an anchor.
If you add the `href` prop to the component it will render as an anchor.

Example:
```js
<Button href="http://www.google.com" target="blank">Open external url</Button>
```

How the component will be rendered:
```js
<a href="http://www.google.com" target="blank">Open external url</a>
```

## Snapshot

Screen Shot 2022-03-17 at 10 23 22

Screen Shot 2022-03-17 at 10 23 31

Screen Shot 2022-03-17 at 10 23 39

## Changes
- Using [styled component "as" ](https://styled-components.com/docs/api#as-polymorphic-prop)to render the the component as an anchor element
- Extending `Link` button from the default `Button` component
- Removing selector by pseudo-class `disabled` because it doesn't work with ancho tag. Replacing it with a conditional
- Add snapshot test to each `Button` type: Default, Link, Outline, and Text
- Add more info to the documentation: Default, Link, Outline, and Text
- Remove `color` primary from the Layout component (used on the documentation). It was affecting any `anchor` tag

### DIFF
--- a/packages/yoga/src/Button/web/Button.jsx
+++ b/packages/yoga/src/Button/web/Button.jsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { func, node, oneOfType, bool } from 'prop-types';
+import { func, node, oneOfType, bool, string } from 'prop-types';
 import StyledButton from './StyledButton';
 
 /** Buttons make common actions more obvious and help users more easily perform them. Buttons use labels and sometimes icons to communicate the action that will occur when the user touches them. */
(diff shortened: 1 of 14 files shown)


frgiovanna (approved):

ericcleao (approved):

mmartinsoliv (approved):

packages/yoga/src/Button/web/Button.jsx
```diff
@@ -13,20 +13,30 @@ const Button = ({
...
-    small={small}
-    secondary={secondary}
-    {...props}
-  >
-    {Icon && <Icon />}
-    {children}
-  </StyledButton>
-);
+}) => {
+  const finalProps = {
```
rafaelcoletagympass, line 17: It's just a suggestion, but if think it's better to create a specific variable to handle the "as" prop instead of creating another object with the props and mutating it.

`
  const ButtonComponent = 'button';
  const LinkComponent = 'a';
 
  let asComponent = ButtonComponent;
  
  if (props.href) asComponent = LinkComponent;
  
  return (
    <StyledButton
      disabled={disabled}
      full={full}
      inverted={inverted}
      onClick={onClick}
      small={small}
      secondary={secondary}
      {...props}
      as="asComponent"
    >
      {Icon && <Icon />}
      {children}
    </StyledButton>
  );
`

PR #451, opened 2022-04-04
Add attendances parameter of Result component as optional

alycecristines:
## Description 📄
This PR changes the attendances parameter of the Result component so that it is optional.
This change arose from the need to use this component in the app's searchMap, and more information about the task can be found [here](https://gympass.atlassian.net/browse/XPLRR-1049).

## Screenshots 📸

| iOS | Android |
| ------ | ----- |
|        |       |

### DIFF
--- a/packages/yoga/src/Result/native/Result.jsx
+++ b/packages/yoga/src/Result/native/Result.jsx
@@ -44,14 +44,12 @@ const Title = styled(Text.Medium)`
   ${({
     theme: {
       yoga: {
-        spacing: { xxxsmall },
         lineHeights: { medium },
       },
     },
   }) => {
     return `
       line-height: ${medium}px;
-      margin-top: ${xxxsmall}px;
     `;
   }}
 `;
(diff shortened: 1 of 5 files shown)


ericcleao (approved):

LidiaYoshino (approved):

packages/yoga/src/Result/native/Result.jsx
```diff
@@ -72,7 +70,7 @@ const Result = ({
   <StyledBox divided={divided} display="flex" flexDirection="row">
     {Avatar && <>{isValidElement(Avatar) ? Avatar : <Avatar />}</>}
     <Content>
-      <Attendances attendances={attendances} rate={rate} />
+      {attendances && <Attendances attendances={attendances} rate={rate} />}
```
leonardokl, line 73: What do you think about supporting an empty list as well? This would be useful in case the attendance list is dynamic

```suggestion
      {!!attendances?.length && <Attendances attendances={attendances} rate={rate} />}
```

leonardokl (approved):

mmartinsoliv (approved):

PR #489, opened 2022-06-07
Updating lerna to v5 and adding nx configs

mmartinsoliv:

## Description 📄

Benchmark before and after with nx

## Lint

| Before | After |
| ------ | ----- |
|image|image|

## Test

| Before | After |
| ------ | ----- |
|image|image|

## Build

| Before | After |
| ------ | ----- |
|image|image|

PS: Nx probably builds package dependencies and lerna does not.

## Cache

There is cache control on behalf of Nx and generating the build command a second time would give a very good number.

image

## Graph

image

Ref: [How Lerna just got 10x faster!](https://blog.nrwl.io/lerna-used-to-walk-now-it-can-fly-eab7a0fe7700)


nx.json
```diff
@@ -0,0 +1,40 @@
+{
+  "extends": "nx/presets/npm.json",
+  "tasksRunnerOptions": {
+    "default": {
+      "runner": "@nrwl/nx-cloud",
+      "options": {
+        "cacheableOperations": ["build", "test", "lint"],
+        "accessToken": "ZTNjOTZlNmQtMjJlYy00NjJkLWEyM2MtNzk2OTllOWJmZmVlfHJlYWQtd3JpdGU="
```
frgiovanna, line 8: Should the token be hard coded here?

frgiovanna (commented):
Nice improvement you have here! I'm impressed with the time decreased on tests and lint, I just left a comment

nx.json
mmartinsoliv: Hi!

nx itself leaves it open too: https://github.com/nrwl/nx/blob/master/nx.json#L17

frgiovanna: Hmm, ok, but do you know if it generates any vulnerability in our project?

mmartinsoliv: Won't bring that kind of problem

dehmirandac2: Checking [the NX doc](https://nx.app/docs/configuring-nx-cloud-cli), they are getting this value from an env. Maybe it could be a better approach

guilhermecardoso-gympass (approved):

guilhermecardoso-gympass:
Great! Congrats

PR #500, opened 2022-06-24
🐛 fix: add color text and opacity for input web

hesugiyama:
## Description 📄

- Add text color and opacity to fix bug on iOS (Safari)
- Update tests snapshots

## Platforms 📲

- [X] Web
- [ ] Mobile

## Type of change 🔍

- [X] Bug fix (non-breaking change which fixes an issue)

### Before 
image

### After 
image

### DIFF
--- a/packages/yoga/src/Input/web/Field.jsx
+++ b/packages/yoga/src/Input/web/Field.jsx
@@ -89,7 +89,7 @@ const Field = styled.input`
       cursor: not-allowed;
       color: ${colors.text.disabled};
       -webkit-text-fill-color: ${colors.text.disabled};
-      opacity: 1; 
+      opacity: 1;
     }
 
     ${placeholder && label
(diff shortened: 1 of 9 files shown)


frgiovanna (approved):

diegomarcuz (approved):

packages/yoga/src/Input/web/Field.jsx
```diff
@@ -88,6 +88,8 @@ const Field = styled.input`
     &:disabled {
       cursor: not-allowed;
       color: ${colors.text.disabled};
+      -webkit-text-fill-color: ${colors.text.disabled};
```
Wendlereis, line 91: As a discovery, could you check if in case you add just `text-fill-color`, styled-component is going to to handle this `-webkit` and other types of fallback.

Letting styled-component handle that will cover other typer of browser engines, not only safari/chrome ones

hesugiyama: image


good catch, it worked fine

ericcleao (approved):

Wendlereis (approved):

guilhermecardoso-gympass (approved):

PR #510, opened 2022-07-12
🚀 feat(icons): add code icon

dpoquet:
## Description 📄

This is a new icon that has been included for the first time in a project but is not yet in the Yoga design system. We have the OK from the Yoga design team to include it in the system.


guilhermecardoso-gympass (approved):

packages/icons/src/svg/code.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="12" height="12" viewBox="0 0 12 12" xmlns="http://www.w3.org/2000/svg">
+<path fill-rule="evenodd" clip-rule="evenodd" d="M14.6708 7.34946C14.8571 6.82955 14.5881 6.25309 14.07 6.0619C13.5518 5.87072 12.9808 6.1372 12.7944 6.65711L8.94812 17.3902C8.76181 17.9101 9.0308 18.4865 9.54894 18.6777C10.0671 18.8689 10.6381 18.6024 10.8245 18.0825L14.6708 7.34946ZM8.18226 7.29693C8.57563 7.6846 8.58026 8.31774 8.1926 8.71111L4.3636 12.5965L8.1926 16.4818C8.58026 16.8752 8.57563 17.5083 8.18226 17.896C7.7889 18.2837 7.15575 18.279 6.76809 17.8857L2.31652 13.3686C1.89449 12.9403 1.89449 12.2526 2.31652 11.8244L6.76809 7.30727C7.15575 6.9139 7.7889 6.90927 8.18226 7.29693ZM15.8196 7.29693C15.4263 7.6846 15.4217 8.31774 15.8093 8.71111L19.6383 12.5965L15.8093 16.4818C15.4217 16.8752 15.4263 17.5083 15.8196 17.896C16.213 18.2837 16.8462 18.279 17.2338 17.8857L21.6854 13.3686C22.1074 12.9403 22.1074 12.2526 21.6854 11.8244L17.2338 7.30727C16.8462 6.9139 16.213 6.90927 15.8196 7.29693Z" fill="white"/>
```
ericcleao, line 2: What happens if we remove the `fill="white"` from the path?

ericcleao: Asking, because all the other icons don't have the fill property set on the icon path

dpoquet: Yes, you're right @ericcleao!! I have removed , thanks.

ericcleao: I asked, because sometimes when we remove the `fill` property the icon breaks, it is caused because it was generated or built in a way that we can´t be used at React Svg (if it is the case is better to talk with the design to show how the icon is looking like and ask to generate another one)

ericcleao: [image](https://github.com/gympass/yoga/pull/510)
vs 
[image](https://github.com/gympass/yoga/pull/507)

dpoquet: Done! I have changed the sizes of the SVG

ericcleao (approved):

caiotracera (approved):

PR #479, opened 2022-05-30
refactor(colors): mofifyed a verve and relax colors and add a new col…

guilhermecardoso-gympass:
…or brandingVerve

[JIRA Issue](https://gympass.atlassian.net/browse/)

## Description 📄

[Just modify a verde and relax colors and add a new color brandingVerve]

## Platforms 📲

- [x] Web
- [x] Mobile

## Type of change 🔍

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How Has This Been Tested? 🧪

[Enter the tips to test this PR]

- [ ] Unit Test
- [x] Snapshot Test

## Checklist: 🔍

- [x] My code follows the contribution guide of this project [Contributing Guide](https://github.com/Gympass/yoga/blob/master/CONTRIBUTING.md)
- [x] Layout matches design prototype: [FIGMA](https://figma.com/file/YOUR_LINK)
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

PR #516, opened 2022-07-19
🐞 fix(colors): change feedback attention dark color

dpoquet:
## Description 📄

I noticed that the `feedback.attention.dark / feedback.attention[1]` hasn't the correct color value according to the Yoga design system https://www.figma.com/file/wn9grDi9mM29Wt7LmQjol7/%E2%9C%85-Yoga-%5BMAIN%5D?node-id=8688%3A10699

The main change was in the` theme.js` file, also, I took advantage to modify the error color in the `Dropdown` component using the right value.

## Platforms 📲

- [x] Web
- [x] Mobile

## Type of change 🔍


ericcleao:
@dpoquet the feedback attention dark color should be `verve` as it was set before. It was a change to improve the contrast including the feedback usage. 
You can check this change here: #479.
If the color is still the old (`brandingVerve`) at the design delivered to you is better to talk with the design team to fix in their side.

ericcleao (changes requested):
@dpoquet the feedback attention dark color should be `verve` as it was set before. It was a change to improve the contrast including the feedback usage. 
You can check this change here: #479.
If the color is still the old (`brandingVerve`) at the design delivered to you is better to talk with the design team to fix in their side.

guilhermecardoso-gympass:
all feedback colors have to used #CA4808 (verve) color more information see on yoga desing system documentation

PR #549, opened 2022-09-06
🚀 Add dropdown to Autocomplete

caiotracera:
[JIRA Issue](https://gympass.atlassian.net/browse/DS-436)

## Description 📄

[Enter the description of your changes]

## Platforms 📲

- [ ] Web
- [ ] Mobile

## Type of change 🔍

This PR adds a dropdown to the AutoComplete component that make it able to show all the suggestions without having to write anything.

- [x] Unit Test
- [x] Snapshot Test

## Checklist: 🔍

- [x] Layout matches design prototype: [FIGMA](https://www.figma.com/file/wn9grDi9mM29Wt7LmQjol7/branch/bsK7IrMaQmToWFvoCMeS6U/%E2%9C%85-Yoga-%5BMAIN%5D)
- [x] I have commented my code, particularly in hard-to-understand areas

## Screenshots 📸

| Before | After |
| ------ | ----- |
|        |       |

### DIFF
--- a/packages/yoga/src/AutoComplete/web/AutoComplete.jsx
+++ b/packages/yoga/src/AutoComplete/web/AutoComplete.jsx
@@ -1,9 +1,10 @@
 /* eslint react/no-array-index-key: 0 */
-import React, { useState } from 'react';
+import React, { useState, useCallback } from 'react';
 import Downshift from 'downshift';
 import { arrayOf, string, func, bool, shape } from 'prop-types';
 import styled from 'styled-components';
 
+import { ChevronDown, ChevronUp } from '@gympass/yoga-icons/src';
 import Input from '../../Input/web/Input';
 
 const escapeRegExp = str => str.replace(/[.*+\-?^${}()|[\]\\]/g, '\\$&');
@@ -158,6 +159,15 @@ const AutoComplete = React.forwardRef(
     ref,
   ) => {
     const [userValue, setUserValue] = useState(value);
+    const [isSuggestionsOpen, setIsSuggestionsOpen] = useState(false);
+
+    const handleOpenSuggestions = useCallback(() => {
+      setIsSuggestionsOpen(true);
+    }, []);
+
+    const handleCloseSuggestions = useCallback(() => {
+      setIsSuggestionsOpen(false);
+    }, []);
 
     return (
       <Downshift
@@ -189,7 +199,7 @@ const AutoComplete = React.forwardRef(
           inputValue,
         }) => {
           const reg = new RegExp(
-            `(${escapeRegExp(inputValue || '').trim() || null})`,
+            `(${escapeRegExp(inputValue || '').trim()})`,
             'gi',
           );
 
@@ -203,14 +213,16 @@ const AutoComplete = React.forwardRef(
             )
             .slice(0, 6);
 
-          const hasSuggestion = isOpen && Boolean(suggestionList.length);
+          if (isOpen && Boolean(suggestionList.length)) {
+            setIsSuggestionsOpen(true);
+          }
 
           return (
             <Wrapper
               className={className}
               style={style}
               full={full}
-              isOpen={hasSuggestion}
+              isOpen={isSuggestionsOpen}
               {...getRootProps()}
               ref={ref}
             >
@@ -224,8 +236,15 @@ const AutoComplete = React.forwardRef(
                 {...getInputProps({
                   onFocus: () => (inputValue.length ? openMenu() : null),
                 })}
+                rightIcon={
+                  isSuggestionsOpen ? (
+                    <ChevronUp onClick={handleCloseSuggestions} />
+                  ) : (
+                    <ChevronDown onClick={handleOpenSuggestions} />
+                  )
+                }
               />
-              {hasSuggestion && (
+              {isSuggestionsOpen && (
                 <List {...getMenuProps()} full={full}>
                   {suggestionList.map((option, optionIndex) => (
                     <Item
@@ -239,7 +258,7 @@ const AutoComplete = React.forwardRef(
                       {option
                         .split(reg)
                         .map((part, index) =>
-                          part.match(reg) ? (
+                          !!inputValue && part.match(reg) ? (
                             <Match key={index}>{part}</Match>
                           ) : (
                             <React.Fragment key={`unmatch-${index}`}>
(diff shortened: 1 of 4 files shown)


ericcleao (approved):

marinamsou (approved):

jonathanbueno (approved):

packages/yoga/src/AutoComplete/web/AutoComplete.jsx
```diff
@@ -200,32 +222,43 @@ const AutoComplete = React.forwardRef(
...
                   onClean(cleanable);
                   clearSelection();
+                  handleCloseSuggestions();
                 }}
                 {...getInputProps({
                   onFocus: () => (inputValue.length ? openMenu() : null),
                 })}
+                rightIcon={
+                  isSuggestionsOpen ? (
+                    <ChevronUp onClick={handleCloseSuggestions} />
```
frgiovanna, line 254: How about adding an aria-label here too? The same you're doing with the `Close` icon

packages/yoga/src/AutoComplete/web/AutoComplete.jsx
```diff
@@ -1,9 +1,10 @@
 /* eslint react/no-array-index-key: 0 */
-import React, { useState } from 'react';
+import React, { useState, useCallback, useRef, useEffect } from 'react';
 import Downshift from 'downshift';
 import { arrayOf, string, func, bool, shape } from 'prop-types';
-import styled from 'styled-components';
+import styled, { css } from 'styled-components';
 
+import { ChevronDown, ChevronUp } from '@gympass/yoga-icons/src';
```
frgiovanna, line 7: Why do we need to get it from the /src?
```suggestion
import { ChevronDown, ChevronUp } from '@gympass/yoga-icons';
```

frgiovanna (approved):

packages/yoga/src/AutoComplete/web/AutoComplete.jsx
caiotracera: Miss it! 😅 Fixed 😁

packages/yoga/src/AutoComplete/web/AutoComplete.jsx
caiotracera: It's an error from the VSCode auto import 😕 fixed!

packages/yoga/src/AutoComplete/web/AutoComplete.jsx
```diff
@@ -200,33 +229,51 @@ const AutoComplete = React.forwardRef(
...
+                rightIcon={
+                  isSuggestionsOpen ? (
+                    <ChevronUp
+                      onClick={handleCloseSuggestions}
+                      aria-label="Close suggestions"
+                    />
+                  ) : (
+                    <ChevronDown
+                      onClick={handleOpenSuggestions}
+                      aria-label="Open suggestions"
```
diegomarcuz, line 269: don't you think it is better to set `aria-label` as props? because gives the possibility of internationalization by the clients

packages/yoga/src/AutoComplete/web/AutoComplete.jsx
```diff
@@ -200,33 +229,51 @@ const AutoComplete = React.forwardRef(
               second.toLowerCase().indexOf(inputValue)
                 ? -1
                 : 1,
-            )
-            .slice(0, 6);
+            );
 
-          const hasSuggestion = isOpen && Boolean(suggestionList.length);
+          if (!!inputValue && isOpen) {
```
diegomarcuz, line 234: As you are checking the truthy value, why not using like this?
```suggestion
          if (inputValue && isOpen) {
```

packages/yoga/src/AutoComplete/web/AutoComplete.jsx
```diff
@@ -200,33 +229,51 @@ const AutoComplete = React.forwardRef(
...
+                      aria-label="Open suggestions"
+                    />
+                  )
+                }
+                ref={inputRef}
               />
-              {hasSuggestion && (
-                <List {...getMenuProps()} full={full}>
+              {isSuggestionsOpen && (
+                <List {...getMenuProps()} full={full} error={!!error}>
```
diegomarcuz, line 276: what do you think?
```suggestion
                <List {...getMenuProps()} full={full} error={Boolean(error)}>
```

packages/yoga/src/AutoComplete/web/AutoComplete.jsx
caiotracera: Hey, @diegomarcuz.
This is just to be sure that the `inputValue` will always be a boolean, to prevent unexpected JavaScript behavior. This is the same as `Boolean(inputValue)`.

packages/yoga/src/AutoComplete/web/AutoComplete.jsx
caiotracera: Hey, @diegomarcuz!
Do you think that the use of the `!!` makes the code harder to read? I don't see a real gain in making that change, but I can do it if you think it improves the readability of the code.

packages/yoga/src/AutoComplete/web/AutoComplete.jsx
caiotracera: Hey, @diegomarcuz!
Great idea! Just pushed the changes for this implementation.

frgiovanna: This is a great point but thinking in all Yoga components, since the DS today doesn't give support to accessibility at all. 
We can talk more about it later, but these accessibility props could be in the System and we can pass it to all components, and also take a look at where we aren't handling it on the component itself.

diegomarcuz: Great idea @frgiovanna. Mainly, because each component could have a specific behavior and requires some accessibility treatment.

packages/yoga/src/AutoComplete/web/AutoComplete.jsx
diegomarcuz: I understood, my comment is just a matter of semantic, but it is fine.

packages/yoga/src/AutoComplete/web/AutoComplete.jsx
diegomarcuz: I think it improves readability, but it is a matter of personal choice, feel free to add.

diegomarcuz (approved):

guilhermecardoso-gympass (approved):

PR #552, opened 2022-09-17
🚀 Add Popover component

caiotracera:
## Description 📄

This PR adds a new component called Popover. 

## Platforms 📲

- [x] Web
- [ ] Mobile

## Type of change 🔍

- [x] Unit Test
- [x] Snapshot Test

## Checklist: 🔍

- [x] Layout matches design prototype: [FIGMA](https://www.figma.com/file/wn9grDi9mM29Wt7LmQjol7/branch/hpkzKVeRDrhagHETO4a5B7/%E2%9C%85-Yoga-%5BMAIN%5D?node-id=29992%3A275548)
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] Any dependent changes have been merged and published in downstream modules

## Screenshots 📸

https://user-images.githubusercontent.com/25802240/191057589-fa09494e-314b-4635-9b66-30601fda7ca4.mp4


YagoAzevedo (approved):

packages/yoga/src/Popover/web/Popover.jsx
```diff
@@ -0,0 +1,192 @@
...
+
+    border-radius: ${radii.small}px;
+    background-color: ${colors.stamina};
+    padding: ${spacing.small}px;
+
+    ${popoverContainerPositionModifier[position]};
+  `}
+`;
+
+export const Title = styled.p`
```
frgiovanna, line 97: Do we need to redeclare the p here since we have the Text component with the same layout?

ericcleao: Agree with, @frgiovanna. We can probably use Text components instead of the Title and Description styled components.

caiotracera: Hey, @ericcleao @frgiovanna and @YagoAzevedo!
Just changed to use the `Text` from Yoga, my bad not seeing this before 😅

Can you review again?

packages/yoga/src/Popover/web/Popover.jsx
```diff
@@ -0,0 +1,71 @@
...
+function Popover({
+  children,
+  title,
+  description,
+  position,
+  width,
+  height,
+  ...props
+}) {
+  const [isPopoverOpen, setIsPopoverOpen] = useState(false);
```
Wendlereis, line 15: We can simplify here 
```suggestion
  const [isOpen, setIsOpen] = useState(false);
```

packages/yoga/src/Popover/web/Popover.jsx
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  function handleHidePopover() {
+    setIsPopoverOpen(false);
+  }
+
+  return (
+    <Wrapper {...props}>
+      {isPopoverOpen && (
+        <PopoverContainer position={position} width={width} height={height}>
```
frgiovanna, line 28: How about adding the `role` [dialog](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role) here? It seems this is the right role for this case, although we also have the [tooltip](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role) one, we can check which one is better but I think we should add a role here, WDYT?

packages/yoga/src/Popover/web/styles.js
```diff
@@ -0,0 +1,109 @@
+import styled, { css } from 'styled-components';
```
frgiovanna, line 1: style files <3

packages/yoga/src/Popover/web/styles.js
```diff
@@ -0,0 +1,109 @@
...
+    },
+  }) => css`
+    margin: 0;
+    margin-bottom: ${spacing.xxxsmall}px;
+  `}
+`;
+
+export const Description = styled(Text.Small)`
+  margin: 0;
+`;
```
Wendlereis, line 109: For me, when we just have one styling property to be added, i would go with the `System` properties. What do you think?

```jsx
<Text.Small m="zero">{value}</Text.Small>
```

packages/doc/content/components/components/popover/index.mdx
```diff
@@ -0,0 +1,27 @@
...
+
+### Usage
+
+```javascript
+<Popover
+  title="Making wellbeing universal"
+  description="Try moving to another spot. Your new favorite activity could be arround the corner."
+  width={265}
+>
+  <Info width={26} height={26} />
```
Wendlereis, line 21: Prefer using the `<icon/>` component

```suggestion
  <Icon as={Info} size="large"  />
```

Wendlereis (commented):
Nice job so far! 

I would like to know if it is completely ready for us to develop as the component spec is not on main branch on figma yet, and i see that is missing some details like the arrow spec and so on.

packages/yoga/src/Popover/web/styles.js
```diff
@@ -0,0 +1,109 @@
...
+export const Title = styled(Text.Medium)`
+  ${({
+    theme: {
+      yoga: { spacing },
+    },
+  }) => css`
+    margin: 0;
+    margin-bottom: ${spacing.xxxsmall}px;
+  `}
+`;
```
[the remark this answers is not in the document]
Wendlereis: I was double checking the Figma file and I notice that TItle should be Body2 Medium which for us would be `Text.Small`

So I would recommend another alternative via JSX using the system properties

```jsx
<Text.Small mb="xxxsmall" fw="medium">{value}<Text.Small>
```

packages/yoga/src/Popover/web/styles.js
Wendlereis: Now i got that you used this margin 0 because you were using `<p>`. So now if the `Text.Small` you can get rid of this styled and use `Text.Small` directly in your jsx without the `m="zero"` prop

caiotracera:
Hey, @Wendlereis!

Firstly, I would like to thank you for the detailed code review. I really appreciate when you guys take some time to review our progress here, and the more comments, the better the final quality of the component.

> I would like to know if it is completely ready for us to develop as the component spec is not on main branch on figma yet, and i see that is missing some details like the arrow spec and so on.

About the Figma, you can double-check with `Georgia Bezerra` or `Edel Marques` at GChat - unfortunally they don't have GitHub accounts. As far as I know, the Figma will still get the final touches, but you will be able to use the component right after the deploy on production that will be announced on our `ds-release` channel at GChat.

guilhermecardoso-gympass (approved):

packages/yoga/src/Popover/web/Popover.jsx
caiotracera: Done!

packages/yoga/src/Popover/web/Popover.jsx
caiotracera: I agree that we can simplify, but I'm used to setting the variables names as descriptive as possible, so there should be not doubt about what's open when reading the code.

WDYT?

packages/yoga/src/Popover/web/styles.js
```diff
@@ -0,0 +1,109 @@
...
+    position,
+    width,
+    height,
+  }) => css`
+    position: absolute;
+
+    width: max-content;
+    max-width: ${width}px;
+    height: max-content;
+    max-height: ${height}px;
```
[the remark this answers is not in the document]
caiotracera: As described [here](https://www.figma.com/file/hpkzKVeRDrhagHETO4a5B7/Popover?node-id=30057%3A277421), the popover container needs to have a dynamic height and width.

But, in some cases, with larger texts, the component can much larger than the expected, so the developer will be able to set the max height and max width of the container.

packages/doc/content/components/components/popover/index.mdx
```diff
@@ -0,0 +1,27 @@
...
+</Popover>
+```
+
```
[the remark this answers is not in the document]
caiotracera: I am working on that!

packages/doc/content/components/components/popover/index.mdx
caiotracera: Done!

packages/yoga/src/Popover/web/styles.js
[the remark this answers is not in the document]
caiotracera: Done!

packages/doc/content/components/components/popover/index.mdx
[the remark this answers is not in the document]
caiotracera: Done!

packages/yoga/src/Popover/web/Popover.jsx
Wendlereis: For me, it is already clear that we are in a component that controls visibility for something, so the `isOpen` is related to it. 

No problem to go with your solution anyways

packages/yoga/src/Popover/web/styles.js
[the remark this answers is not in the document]
Wendlereis: As far I understand from figma, the user shouldn't change the contract for the max-width size.

packages/doc/content/components/components/popover/index.mdx
```diff
@@ -0,0 +1,128 @@
...
+metaDescription: 'Popover Component'
+---
+
+# Popover
+
+### Reference
+
+Gympass `<Popover />` description
+
+### Usage
```
Wendlereis, line 13: I still would split the basic usage as you did in last commit, and another section explaining the position prop and its results as you did here

like [Mui does](https://mui.com/material-ui/react-tooltip/#positioned-tooltips)

packages/doc/content/components/components/popover/index.mdx
```diff
@@ -0,0 +1,128 @@
...
+
+```javascript
+<div style={{display: 'flex', alignItems: 'center', justifyContent: 'space-between', width: '100%', flexDirection: 'column', margin: '0 15rem 0', gap: '2rem'}}>
```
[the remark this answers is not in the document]
caiotracera: Done!

packages/doc/content/components/components/popover/index.mdx
caiotracera: Done!

PR #618, opened 2023-02-01
💣 fix(Icon): update toiletries name

diegomarcuz:
## Description 📄

This update the Toiletries icon name

## Platforms 📲

- [X] Web
- [X] Mobile

## Type of change 🔍

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist: 🔍

## Screenshots 📸

| Before | After |
| ------ | ----- |
|        |       |


diogoverde (approved):

ericcleao (approved):

marinamsou (approved):

wellingtonnascimento (approved):

packages/icons/src/ui.js
```diff
@@ -303,7 +303,7 @@ export {
   Time,
   TimeFilled,
   Token,
-  Tolietries,
```
caiotracera, line 306: Is there a chance this change will break some application that is using this icon with the wrong name?

diegomarcuz: good point that's why we are trying to address it as a breaking change

diegomarcuz: what do you think?

caiotracera: I like the approach of addressing this as a breaking change. In addition, I would also suggest to keep two exports: one with the wrong name (just like it is right now), and another one with your fix.

```jsx
export {
...
Tolietries,
Toiletries: Tolietries,
....
}
```

That way, we allow teams time to fix the problem before it happens. That way, we allow teams time to fix the problem before it happens. And therefore, we have set a deadline for the incorrect version to be removed, such as 30 days. WDYT?

CC @matheushdsbr @caioalexandrebr

PR #634, opened 2023-03-23
🚀 feat(Input): add a11y attributes to Input

marcos-creuz:
[JIRA Issue](https://gympass.atlassian.net/browse/ITK-469)

## Description 📄

This PR adds some accessibility improvements to the Input component, focusing mainly in improving the usability with screen readers:

- Adds two new props to the Input component: `a11yId` and `includeAriaAttributes` (more details about these below);
- Adds `aria-labelledby` to the input element, so that screen readers can announce the label of the input. If no `a11yId` is provided, `aria-label` is added instead (`aria-labelledby` is preferable when there is a visible text that labels an element according to [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label));
- Adds `aria-describedby` to the input element, referencing the helper as it's description element. This allows the screen reader to announce helper and error messages associated with the input when the user focuses the input (works only if the `a11yId` prop is provided).
- Adds `aria-live="polite"` to the Helper component, so that error messages can be announced by the screen reader as soon as they appear;
- Adds `aria-invalid` to the input element, setting it's value to `true` when the `error` prop has a truthy value. This allows the screen reader to announce that the input has invalid data.

#### Generating HTML IDs for a11y
In order to use `aria-labelledby` and `aria-describedby` on the input element, HTML IDs had to be added to the label and helper elements. Some approaches were considered in order to do this (thanks @caiotracera for giving some ideas here):

1. The optimal solution would be using React's [useId hook](https://beta.reactjs.org/reference/react/useId), made specifically to solve this problem. Unfortunately, it is only available on React 18 (Yoga is currently on React 17.0.2).
2. Generate a UUID for the input, store it with `useRef` or `useState`, and use it as a prefix for the HTML IDs. The pro of this approach is that `aria-labelledby` and `aria-describedby` would automatically work on all inputs without the need of passing a prop for that. The con is that some inconsistencies might happen with server side rendering (I honestly don't have experience with SSR and would like to know your opinions on this).
3. Manually pass an ID as a prop to the Input component, which seemed to be the safest approach to start with. Although this approach is not scalable, since devs need to manually pass an ID to each input that they want to have `aria-labelledby` and `aria-describedby` working, it at least makes it possible to do it somehow. Another downside of this approach is that making the ID prop (`a11yId`) optional made the code a bit hard to read due to, for example, conditional props passing.

More about this can be found in this [SO question](https://stackoverflow.com/questions/29420835/how-to-generate-unique-ids-for-form-labels-in-react) and in this [React issue](https://github.com/facebook/react/issues/5867).

#### New `includeAriaAttributes` prop
Besides the new `a11yId` prop, this PR also adds another prop to the Input component: `includeAriaAttributes`. This prop has a `true` value by default and was added to give the possibility of including or not ARIA attributes on the Input, depending on the situation.

The need for this prop arose from components that extend the Input component and have their own ARIA implementations, like Dropdown and AutoComplete (by the way, the usability of these two components with screen readers also need to be revisited in the future, even though they have the ARIA attributes added by downshift).

## Platforms 📲

- [x] Web
- [ ] Mobile

## Type of change 🔍

- Open the Input page in the documentation app;
- Start running a screen reader (on mac, press cmd+F5 to start VoiceOver);
- Focus the input element:
  - it should announce something like "Find an activity, edit text" (the label is announced due to `aria-label`, since no `a11yId` has being passed yet);
- Press the `< >` button in order to edit the code;
- Edit the code adding `error={value == "error" && "Error message"}` to the list of props being passed to the `Input` component;
- Focus the input and type the word "error":
  - it should announce "Error message" as soon as the input enters in error state (due to `aria-live`);
- Focus any other element and then focus the input element again (still in the error state):
  - it should mention "invalid data" during the announcement (due to `aria-invalid`). Although, in this case it's not expected to have the error description announced, since there's no link between the helper and the input yet;
- Edit the code once again, now adding `a11yId="id"` to the list of props being passed to the `Input` component;
- Focus the input element:
  - it should now announce the content of the helper component too (i.e. the error message if the the input has errors or the helper text otherwise). That works now due to `aria-describedby`, that could be added because of the `a11yId`;
- Edit the code one last time, adding now `includeAriaAttributes={false}` to the list of props being passed to the `Input` component (feel free to keep the other added props or remove them);
- Focus the input element:
  - it should announce "edit text, blank". The label of the input is not expected to be announced due to the removal of `aria-label`/`aria-labelledby`.

- [ ] Unit Test
- [x] Snapshot Test

## Checklist: 🔍

## Screenshots 📸

The videos below do not have audio, but the announcements of VoiceOver are shown as text in the gray box.

Before:

https://user-images.githubusercontent.com/10004004/226908572-03c37f7d-79d3-419e-aacf-3ff60558c58a.mov

- The label of the input is not announced
- The screen reader is not aware whether the input has invalid data or not
- The helper text and the error description are not announced

After:

https://user-images.githubusercontent.com/10004004/226908725-d87f0b9f-e846-4b90-a929-0a2facfa709e.mov

- The label of the input is announced (even when no `a11yId` is passed)
- The screen reader announces if the input has invalid data
- The error description is announced at least once due to `aria-live`, or everytime the input is focused when an `a11yId` is provided

### DIFF
--- a/packages/yoga/src/AutoComplete/web/AutoComplete.jsx
+++ b/packages/yoga/src/AutoComplete/web/AutoComplete.jsx
@@ -273,6 +273,7 @@ const AutoComplete = React.forwardRef(
                   )
                 }
                 ref={inputRef}
+                includeAriaAttributes={false}
               />
               {isSuggestionsOpen && (
                 <List {...getMenuProps()} full={full} error={!!error}>
--- a/packages/yoga/src/Dropdown/web/Dropdown.jsx
+++ b/packages/yoga/src/Dropdown/web/Dropdown.jsx
@@ -305,6 +305,7 @@ const Dropdown = React.forwardRef(
                 label={label}
                 full={full}
                 ref={inputRef}
+                includeAriaAttributes={false}
                 {...getInputProps()}
               />
               <Button
--- a/packages/yoga/src/Dropdown/web/__snapshots__/Dropdown.test.jsx.snap
+++ b/packages/yoga/src/Dropdown/web/__snapshots__/Dropdown.test.jsx.snap
@@ -1548,6 +1548,7 @@ exports[`<Dropdown /> should match snapshot with error 1`] = `
       class="c11"
     >
       <span
+        aria-live="polite"
         class="c12"
       >
         Please, select one activity
--- a/packages/yoga/src/Input/web/Helper.jsx
+++ b/packages/yoga/src/Input/web/Helper.jsx
@@ -47,9 +47,14 @@ const Helper = ({
   maxLength,
   length,
   hideMaxLength,
+  a11yId,
 }) => (
   <Wrapper disabled={disabled} error={error}>
-    {(error || helper) && <Info as="span">{error || helper}</Info>}
+    {(error || helper) && (
+      <Info as="span" aria-live="polite" {...(a11yId && { id: a11yId })}>
+        {error || helper}
+      </Info>
+    )}
     {maxLength && !hideMaxLength && (
       <Info as="span">
         {length}/{maxLength}
@@ -65,6 +70,7 @@ Helper.propTypes = {
   maxLength: number,
   length: number,
   hideMaxLength: bool,
+  a11yId: string,
 };
 
 Helper.defaultProps = {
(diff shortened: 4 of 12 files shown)


brunoeris (approved):

MicaelRodrigues (approved):

matheushdsbr:
@all-contributors please add @marcos-creuz for code

PR #669, opened 2023-07-12
🚀 feat(tokens|common): add typescript

matheushdsbr:

## Description 🧘🏻‍♀️

In this PR was added **Typescript** in `tokens` and `common` packages.

## Platforms 📲

- [ ] Web
- [ ] Mobile

## Type of change 🔍

- [ ] Unit Test
- [ ] Snapshot Test

## Checklist: 🔍

## Video 📸

https://github.com/gympass/yoga/assets/32910717/208e3404-4340-47c4-9d12-f5d273001bfe


packages/common/src/elevate.ts
```diff
@@ -33,21 +45,33 @@ function createShadow({ level, color, depth, spread }) {
...
     createShadow({ level: 1, color, depth, spread }),
     createShadow({ level: 2, color, depth, spread }),
   ];
 
-  return all[level] || all;
+  if (level) {
+    return all[level];
+  }
+
+  return all;
```
MicaelRodrigues, line 74: The previous way seemed more clear, i would only using nullish coalescing operator. WDYT?
```suggestion
  all[level] ?? all;
```

packages/tokens/src/global/borders.ts
```diff
@@ -0,0 +1,11 @@
+export interface BorderProps extends Array<number> {
```
MicaelRodrigues, line 1: Optional, but i think the TS community prefers `<type>[]` [syntax](https://www.typescripttutorial.net/typescript-tutorial/typescript-array-type/#:~:text=In%20TypeScript%2C%20an%20array%20is,let%20arr%3A%20type%5B%5D%20syntax.), to avoid name conflict with [JS Array class](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array).
```suggestion
export interface BorderProps extends number[] {
```

packages/tokens/src/global/breakpoints.ts
```diff
@@ -0,0 +1,101 @@
...
+    margin: number;
+    gutter: number;
+  };
+
+  xxxl: {
+    width: number;
+    margin: number;
+    gutter: number;
+  };
+}
```
MicaelRodrigues, line 49:
```suggestion
type BreakpointType = 'xxs' | 'xs' | 'sm' | 'md' | 'lg' | 'xl' | 'xxl' | 'xxxl';

export interface Breakpoint {
  width: number;
  margin: number;
  gutter: number;
}

export interface BreakpointsProps {
  [key: BreakpointType]: Breakpoint;
}
```

packages/tokens/src/global/font-sizes.ts
```diff
@@ -0,0 +1,27 @@
...
+  fontSizes.small,
+  fontSizes.medium,
+  fontSizes.large,
+  fontSizes.xlarge,
+  fontSizes.xxlarge,
+  fontSizes.xxxlarge,
+  fontSizes.huge,
+] = fontSizes;
+
+export default fontSizes;
```
MicaelRodrigues, line 27: Why not?
```suggestion
const fontSizes = {
  xxsmall: 10,
  xsmall: 12,
  small: 14,
  medium: 16,
  large: 20,
  xlarge: 24,
  xxlarge: 32,
  xxxlarge: 40,
  huge: 48,
};

export type FontSizesProps = typeof fontSizes;
```

The same suggestion for other approaches.

packages/tokens/src/global/font-weights.ts
```diff
@@ -0,0 +1,19 @@
...
+
+[
+  fontWeights.light,
+  fontWeights.regular,
+  fontWeights.medium,
+  fontWeights.bold,
+  fontWeights.black,
+] = fontWeights;
+
+export default fontWeights;
```
MicaelRodrigues, line 19: Same as above

packages/tokens/src/global/line-heights.ts
```diff
@@ -0,0 +1,27 @@
...
+  lineHeight.small,
+  lineHeight.medium,
+  lineHeight.large,
+  lineHeight.xlarge,
+  lineHeight.xxlarge,
+  lineHeight.xxxlarge,
+  lineHeight.huge,
+] = lineHeight;
+
+export default lineHeight;
```
MicaelRodrigues, line 27: Same as above

packages/tokens/src/global/breakpoints.ts
rafaelcoletagympass: Great suggestion!

packages/common/src/elevate.ts
caiotracera: Hey, @MicaelRodrigues!
I don't think the previous way is more clear. Someone new to JS may have difficulty to understand `all[level] ?? all`. Contrary to what you said, I believe that this new implementation is easier to read and to understand the code.

Don't you think it's easier to read the statements if we were reading like it was an English text, instead of a code full of JS patterns and operators?

packages/tokens/src/global/breakpoints.ts
```diff
@@ -0,0 +1,101 @@
...
+    margin: 71,
+    gutter: 24,
+  },
+
+  xxxl: {
+    width: 1600,
+    margin: 71,
+    gutter: 24,
+  },
+};
```
caiotracera, line 99:
```suggestion
} as const;
```

So we can make the IntelliSense and the Editor/autocomplete smarter

packages/tokens/src/global/colors.ts
```diff
@@ -1,116 +1,32 @@
...
- * @type {Color}
- */
 const clear = '#F5F5FA';
-
-/**
- * @type {Color}
- */
 const white = '#FFFFFF';
 
 const colors = {
```
caiotracera, line 32:
```ts
export default colors = {
  vibin,
  hope,
  energy,
  relax,
  peace,
  brandingVerve,
  verve,
  uplift,
  deepPurple,
  yoga,
  success,
  neutral,
  attention,
  stamina,
  deep,
  medium,
  light,
  clear,
  white,
} as const;
```

packages/tokens/src/global/elevations.android.ts
```diff
@@ -0,0 +1,13 @@
+export interface ElevationsProps extends Array<number> {
```
caiotracera, line 1: Same here, regarding the @MicaelRodrigues' optional suggestion
```suggestion
export interface ElevationsProps extends number[] {
```

packages/tokens/src/global/borders.ts
goncalo-matos: I agree (it's also more consistent with what our IDEs show us). I think there are eslint rules to autofix this 🤔

packages/tokens/src/global/font-sizes.ts
caiotracera: In addition to @MicaelRodrigues suggestion, I would add the `as const` at the end of the `fontSizes` object, like this:


```ts
const fontSizes = {
	  xxsmall: 10,
	  xsmall: 12,
	  small: 14,
	  medium: 16,
	  large: 20,
	  xlarge: 24,
	  xxlarge: 32,
	  xxxlarge: 40,
	  huge: 48,
	} as const;
```

packages/tokens/src/global/radii.ts
```diff
@@ -0,0 +1,13 @@
...
+  small?: number;
+  regular?: number;
+  circle?: number;
+}
+
+const radii: LineHeightProps = [0, 4, 8, 16, 9999];
+
+[radii.sharp, radii.xsmall, radii.small, radii.regular, radii.circle] = radii;
+
+export default radii;
```
caiotracera, line 13: I believe you should do the same here regarding @MicaelRodrigues and my comments <3

packages/tokens/src/global/spacing.ts
```diff
@@ -0,0 +1,33 @@
...
+  spacing.medium,
+  spacing.large,
+  spacing.xlarge,
+  spacing.xxlarge,
+  spacing.xxxlarge,
+  spacing.huge,
+  spacing.xhuge,
+] = spacing;
+
+export default spacing;
```
caiotracera, line 33: I believe you should do the same here regarding @MicaelRodrigues and my comments <3

packages/common/package.json
```diff
@@ -4,11 +4,13 @@
...
   "homepage": "https://github.com/Gympass/yoga",
   "license": "MIT",
-  "main": "src/index.js",
+  "main": "src/index.ts",
+  "types": "src/index.ts",
   "scripts": {
-    "build": "yarn build:cjs && yarn build:esm",
-    "build:cjs": "NODE_ENV=cjs babel ./src --out-dir dist/cjs",
-    "build:esm": "NODE_ENV=esm babel ./src --out-dir dist/esm",
+    "build": "yarn build:types && yarn build:cjs && yarn build:esm",
```
goncalo-matos, line 10: **suggestion**: wdyt of parallelizing this? Something like https://www.npmjs.com/package/npm-run-all can be helpfull

packages/common/src/elevate.ts
```diff
@@ -33,21 +45,33 @@ function createShadow({ level, color, depth, spread }) {
   return shadows.slice(0, depth).join();
 }
 
-function elevate({
+interface ElevateProps {
+  color?: string;
+  level?: number;
```
goncalo-matos, line 50: **suggestion**: since this only supports up to 3, shouldn't our types reflect this?

quick and dirty way would be to do this (we could build a Range util later):
```suggestion
  level?: 0 | 1 | 2 | 3 
```

packages/common/src/elevate.ts
goncalo-matos: honestly, I understand both points but there's two things that seem more important to me here:
- what if I give it a higher level? Right now, as we accept any number I could give it a 500 and would get an undefined. Should we create the array based on the given level?
- this function signature is, IMO, weird. It can either return an array or a value. I'm not the biggest fan of doing this as it makes the usage a lot more complex. Maybe we should type the return differently based on the fact that a level is given or not using [function overloads](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads)

packages/tokens/src/global/borders.ts
```diff
@@ -0,0 +1,11 @@
...
+  zero?: number;
+  small?: number;
+  medium?: number;
+}
+
+const border: BorderProps = [0, 1, 2];
+
+[border.zero, border.small, border.medium] = border;
```
goncalo-matos, line 9: **suggestion**: I think we can remove this line. I'm not really following what it's doing 🤔

packages/tokens/src/global/breakpoints.ts
```diff
@@ -0,0 +1,101 @@
+export interface BreakpointsProps {
```
goncalo-matos, line 1: **suggestion**: these aren't really props. How about just Breakpoints?
```suggestion
export interface Breakpoints {
```

packages/tokens/src/global/colors.ts
goncalo-matos: For the future, not for now, we should probably avoid exporting default. we're polluting the bundle with all of the colors if we only use one

(named exports scale a lot better than default exports :P)

packages/tokens/src/global/elevations.ios.ts
```diff
@@ -0,0 +1,21 @@
...
+
+const elevateOptions = {
+  color: colors.medium,
+  depth: 1,
+  spread: false,
+  fallback: true,
+};
+
+const [zero, small, medium, large] =
+  elevate(elevateOptions) as [string, string, string, string];
```
goncalo-matos, line 12: **question**: `as` casts hide potential errors (it's us saying we know better than typescript). Do we really need this?

tsconfig.json
```diff
@@ -0,0 +1,27 @@
+{
+  "compilerOptions": {
+    "target": "es2016",
```
goncalo-matos, line 3: **question**: could we make this a bit more recent? :P

tsconfig.json
```diff
@@ -0,0 +1,27 @@
...
+      "@gympass/yoga-tokens/*": ["./packages/tokens/src/*"],
+      "@gympass/yoga-tokens": ["./packages/tokens/src"]
+    },
+    "declaration": true,
+    "outDir": "./dist/",
+    "esModuleInterop": true,
+    "forceConsistentCasingInFileNames": true,
+    "allowSyntheticDefaultImports": true,
+    "strict": true,
+    "noImplicitAny": false,
```
goncalo-matos, line 18: one day, I'd love to have this as true :D

packages/common/src/elevate.ts
MicaelRodrigues: > Hey, @MicaelRodrigues! I don't think the previous way is more clear. Someone new to JS may have difficulty to understand `all[level] ?? all`. Contrary to what you said, I believe that this new implementation is easier to read and to understand the code.
> 
> Don't you think it's easier to read the statements if we were reading like it was an English text, instead of a code full of JS patterns and operators?

> * could give it a 500 and would get an undefined. Should we create the arr

It's a question of flavour :) We expect people working in this code to know javascript fundamentals. If they don't we are here to help improve the code... :)
I agree with you in "as it was an English text" and both options read plain english to me (ex: my suggestion reads ("a" or "b"). 

Anyway, the important concerns in this function are the ones @goncalo-matos exposed ;)

packages/tokens/src/global/breakpoints.ts
MicaelRodrigues: Maybe we don't need to `export interface Breakpoint`

packages/common/src/elevate.ts
leonardokl: Can we use the value "0" as `level`? If so the current implementation will ignore the level when it is "0".

tsconfig.json
MicaelRodrigues: Fuel to the fire :P
https://www.freecodecamp.org/news/javascript-new-features-es2020/

matheushdsbr: For sure!

But I think it's better to do this update at another time, as there would be a lot to validate in this PR

packages/tokens/src/global/borders.ts
matheushdsbr: Initially I thought of doing it this way, but when I change it to `number[]` I get this error:
"An interface can only extend a qualified identifier/name with optional type arguments"

Googling I understood that this error is because we can only extend an identifier or a qualified name with optional type arguments when defining an interface, in this case `number[]` is an array of numbers, not an identifier or a qualified name and that's why the error happens.

So what I understand is that in this case I should use `Array<number>`, but it still makes me confused 🤔

packages/common/src/elevate.ts
```diff
@@ -33,21 +45,33 @@ function createShadow({ level, color, depth, spread }) {
...
+  depth?: number;
+  spread?: boolean;
+  fallback?: boolean;
+}
+
+function  elevate({
```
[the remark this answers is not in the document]
matheushdsbr: Great suggestion!! @leonardokl 🚀🚀

matheushdsbr:
@MicaelRodrigues @goncalo-matos @leonardokl @caiotracera 

Hey guys!

When changing the structure of the tokens by removing the arrays, this test broke: `/packages/yoga/src/Theme/helpers/themeReader/base.test.js` on line 44.

Investigating the `theme` (being imported on line 3) I realized that we have the `createFakeObj()` function and if we change the structure of the export with the arrays it will not work well and will generate breakchanges.

I believe that it would be better, at least for now, to continue with the old structure (with arrays) and in the future to investigate solutions to remove this structure and update with suggestions that you commented here.

I believe I wasn't very clear, I still don't quite understand how this function works.

I would like your opinion on the case!

MicaelRodrigues:
> @MicaelRodrigues @goncalo-matos @leonardokl @caiotracera
> 
> Hey guys!
> 
> When changing the structure of the tokens by removing the arrays, this test broke: `/packages/yoga/src/Theme/helpers/themeReader/base.test.js` on line 44.
> 
> Investigating the `theme` (being imported on line 3) I realized that we have the `createFakeObj()` function and if we change the structure of the export with the arrays it will not work well and will generate breakchanges.
> 
> I believe that it would be better, at least for now, to continue with the old structure (with arrays) and in the future to investigate solutions to remove this structure and update with suggestions that you commented here.
> 
> I believe I wasn't very clear, I still don't quite understand how this function works.
> 
> I would like your opinion on the case!



> packages/tokens/src/global/font-sizes.ts

I see, we need to keep "array" as the return in order to prevent a breaking change.
So, WDYT of using `Object.values`? Example:
```

const fontSizes = {
   xxsmall: 10,
   xsmall: 20
}

export default Object.values(fontSizes);
```

matheushdsbr:
Great suggestion @MicaelRodrigues!! 🚀🚀

But when testing I found a problem, the old structure [with array] allows us to use tokens in two ways:

`tokens.fontSizes[0]` or `tokens.fontSizes.xxsmall`

All consoles were made in this test file:

```
  packages/yoga/src/Theme/helpers/themeReader/base.test.js

  console.log(theme.fontSizes);
```

Consoling the old structure with arrays:
```
[Function: ProxyPolyfill] {
  '0': [Getter/Setter],
  '1': [Getter/Setter],
  '2': [Getter/Setter],
  xxsmall: [Getter/Setter],
  xsmall: [Getter/Setter],
  small: [Getter/Setter],
}
```

With your suggestion we can use just like:

`tokens.fontSizes.xxsmall`

Code

```
const sizes = {
  xxsmall: 10,
  xsmall: 20,
  small: 30
}

const fontSizes = Object.values(sizes);
```

Result

```
[Function: ProxyPolyfill] {
  '0': [Getter/Setter],
  '1': [Getter/Setter],
  '2': [Getter/Setter],
}
```

But!!! Your suggestion made me think better and I thought of this way:

```
const sizes = {
  xxsmall: 10,
  xsmall: 20,
  small: 30
}

const fontSizes = {...Object.values(sizes), ...sizes};
```

Result:

```
[Function: ProxyPolyfill] {
  '0': [Getter/Setter],
  '1': [Getter/Setter],
  '2': [Getter/Setter],
  xxsmall: [Getter/Setter],
  xsmall: [Getter/Setter],
  small: [Getter/Setter],
}
```

With that way we can continue using the same tokens as in the array structure:

`tokens.fontSizes[0]` or `tokens.fontSizes.xxsmall`

\o/

tsconfig.json
MicaelRodrigues: @matheushdsbr Pls don't forget to open a new PR with this :)

MicaelRodrigues (approved):

caiotracera (approved):

rafaelcoletagympass (approved):

leonardokl (approved):

PR #643, opened 2023-03-29
🐞 fix(Popover): prevent event bubbling to parent

MicaelRodrigues:
[JIRA Issue](https://gympass.atlassian.net/browse/PPE-74)

## Description 📄

This PR updates the popover to make sure events are not bubbled to parent. Example: if you press the popover inside a `anchor`, you want to make sure the link is not opened and instead the popover is visible.

I've update the popover trigger to be a `button`. Here's why: https://www.builder.io/blog/buttons#the-problem-with-divs

I've also added [onTouchStart](https://www.w3schools.com/jsref/event_touchstart.asp) and [onTouchEnd](https://www.w3schools.com/jsref/event_touchend.asp) events to provide better mobile support.

## Platforms 📲

- [x] Web
- [ ] Mobile

## Type of change 🔍

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How Has This Been Tested? 🧪

[Enter the tips to test this PR]

- [ ] Unit Test
- [x] Snapshot Test

## Checklist: 🔍

- [x] My code follows the contribution guide of this project [Contributing Guide](https://github.com/Gympass/yoga/blob/master/CONTRIBUTING.md)
- [x] Layout matches design prototype: [FIGMA](https://figma.com/file/YOUR_LINK)
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

### (Mobile) Before

https://user-images.githubusercontent.com/796443/228497653-228c67b9-dd6a-4a0a-9412-c12da48ed614.mov

### (Mobile) After

https://user-images.githubusercontent.com/796443/228497782-a8cb713c-27ad-4bda-bf0c-b93a9d78cf02.mov

### (Desktop) Before

https://user-images.githubusercontent.com/796443/228498095-800d7d1d-d2e0-4097-8cc9-1642e4cbfe3b.mov

### (Desktop) after

https://user-images.githubusercontent.com/796443/228498167-761108bc-f7ee-452e-9005-d55994920c6d.mov

PR #678, opened 2023-08-22
fix: fix popover behavior to accept click (default) OR hover event

flavia-moraes:
[issue JIRA](https://gympass.atlassian.net/browse/)

## 🐞[FIX]: Popover behavior for mobile

To adjust the behavior for mobile, the onTouch event was removed and the onClick event was preserved to keep the popover open without the need to keep pressing the button.

For this, the click event will be considered the default and a listener has been added for clicks outside the button element. The hover event has been kept as optional, if chosen the click event should be overridden for this behavior.
Keyboard event by default was also kept with click event.
## Platforms 📲

- [x] Web
- [ ] Mobile

## Type of change 🔍

- [x] Unit Test
- [x] Snapshot Test

## Checklist: 🔍

## Screenshots 📸

**[BEFORE]**

Bug report video:

https://github.com/gympass/yoga/assets/141175689/d6d39bb0-6958-4a16-9a1d-1b5d8cf1871e

**[AFTER]**

Mobile, after fix with click behavior

https://github.com/gympass/yoga/assets/141175689/2714f61b-0ea0-44e0-84c6-5381ef85d5a3

Mobile, after fix with click behavior and keyboard

https://github.com/gympass/yoga/assets/141175689/c563ddff-a058-49a3-8747-2bc29f3d6a8f

Web, after fix with click and keboard

https://github.com/gympass/yoga/assets/141175689/64f5a6ad-4d8e-431a-a9d2-219e6ce15194

Web, after fix with hover option

https://github.com/gympass/yoga/assets/141175689/7d99f20a-0bd3-43ff-94c0-f0e519fd202a

### DIFF
--- a/packages/yoga/src/Popover/web/Popover.jsx
+++ b/packages/yoga/src/Popover/web/Popover.jsx
@@ -1,5 +1,5 @@
-import React, { useState } from 'react';
-import { node, number, oneOf, string } from 'prop-types';
+import React, { useState, useEffect, useRef } from 'react';
+import { bool, node, number, oneOf, string } from 'prop-types';
 import { Text } from '@gympass/yoga';
 
 import { PopoverContainer, PopoverButton, Wrapper } from './styles';
@@ -13,9 +13,25 @@ function Popover({
   height,
   zIndex,
   a11yId,
+  hover,
   ...props
 }) {
+  const ref = useRef(null);
   const [isPopoverOpen, setIsPopoverOpen] = useState(false);
+
+  useEffect(() => {
+    const handleClickOutside = event => {
+      if (ref.current && !ref.current.contains(event.target)) {
+        setIsPopoverOpen(false);
+      }
+    };
+
+    document.addEventListener('click', handleClickOutside, true);
+
+    return () => {
+      document.removeEventListener('click', handleClickOutside, true);
+    };
+  }, [setIsPopoverOpen]);
 
   const handleShowPopover = () => {
     setIsPopoverOpen(true);
@@ -23,6 +39,10 @@ function Popover({
 
   const handleHidePopover = () => {
     setIsPopoverOpen(false);
+  };
+
+  const toggleOpenPopover = () => {
+    setIsPopoverOpen(!isPopoverOpen);
   };
 
   return (
@@ -49,16 +69,11 @@ function Popover({
 
       <PopoverButton
         {...(a11yId && { 'aria-describedby': a11yId })}
-        onMouseEnter={handleShowPopover}
-        onMouseLeave={handleHidePopover}
-        onTouchStart={handleShowPopover}
-        onTouchEnd={handleHidePopover}
-        onClick={event => event.preventDefault()}
-        onKeyDown={event => {
-          if (event.key === 'Enter') {
-            setIsPopoverOpen(current => !current);
-          }
-        }}
+        ref={ref}
+        onMouseEnter={hover ? handleShowPopover : undefined}
+        onMouseLeave={hover ? handleHidePopover : undefined}
+        onClick={!hover ? toggleOpenPopover : undefined}
+        onKeyDown={event => event.key === 'Enter' && toggleOpenPopover}
       >
         {children}
       </PopoverButton>
@@ -89,6 +104,7 @@ Popover.propTypes = {
   width: number,
   height: number,
   zIndex: number,
+  hover: bool,
 };
 
 Popover.defaultProps = {
@@ -98,6 +114,7 @@ Popover.defaultProps = {
   width: 265,
   height: 200,
   zIndex: 1,
+  hover: false,
 };
 
 export default Popover;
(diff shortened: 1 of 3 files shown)


packages/yoga/src/Popover/web/Popover.jsx
```diff
@@ -13,10 +13,26 @@ function Popover({
   height,
   zIndex,
   a11yId,
+  hover,
   ...props
 }) {
+  const ref = useRef(null);
```
MicaelRodrigues, line 19: Make the component a forwardRef, so you cam also receive external refs. WDYT?
(Example in dialog)
`    const popoverRef = useCombinedRefs(forwardedRef);`

packages/yoga/src/Popover/web/Popover.jsx
```diff
@@ -49,16 +69,11 @@ function Popover({
...
-        onClick={event => event.preventDefault()}
-        onKeyDown={event => {
-          if (event.key === 'Enter') {
-            setIsPopoverOpen(current => !current);
-          }
-        }}
+        ref={ref}
+        onMouseEnter={hover ? handleShowPopover : undefined}
+        onMouseLeave={hover ? handleHidePopover : undefined}
+        onClick={!hover ? toggleOpenPopover : undefined}
```
MicaelRodrigues, line 75: As discussed offline, previous behaviour should be preserved:

https://github.com/gympass/yoga/pull/643

packages/yoga/src/Popover/web/Popover.jsx
```diff
@@ -13,10 +13,26 @@ function Popover({
   height,
   zIndex,
   a11yId,
+  hover,
```
MicaelRodrigues, line 16: How will you control the value of this prop?

Desktop should keep the `hover` behaviour right?

packages/yoga/src/Popover/web/Popover.jsx
```diff
@@ -25,6 +41,10 @@ function Popover({
     setIsPopoverOpen(false);
   };
 
+  const toggleOpenPopover = () => {
+    setIsPopoverOpen(!isPopoverOpen);
+  };
```
MicaelRodrigues, line 46: This is a bad practise. Use[ setState callback](https://carlogino.com/blog/react-setstate-callback-function) :)

```suggestion
  const toggleOpenPopover = () => {
      setIsPopoverOpen(current => !current);
  };
```

packages/yoga/src/Popover/web/Popover.jsx
```diff
@@ -49,16 +69,11 @@ function Popover({
...
-        onKeyDown={event => {
-          if (event.key === 'Enter') {
-            setIsPopoverOpen(current => !current);
-          }
-        }}
+        ref={ref}
+        onMouseEnter={hover ? handleShowPopover : undefined}
+        onMouseLeave={hover ? handleHidePopover : undefined}
+        onClick={!hover ? toggleOpenPopover : undefined}
+        onKeyDown={event => event.key === 'Enter' && toggleOpenPopover}
```
MicaelRodrigues, line 76: toggleOpenPopover should be called. Also keep the code as before:
```suggestion
        onKeyDown={event => {
          if (event.key === 'Enter') {
             toggleOpenPopover();
         }
        }}
```

Becase if `event.key === 'Enter'` is false, the function will return `false`. But it should be `void`

packages/yoga/src/Popover/web/Popover.jsx
MicaelRodrigues: > and click will default

I would agree with you, but that would break existing implementations that are relying on hover - which was the original behavior. So `hover` should be `true`.

Agree? :)

packages/yoga/src/Popover/web/Popover.jsx
MicaelRodrigues: @flavia-moraes don't forget this change please :)

packages/yoga/src/Popover/web/Popover.jsx
flavia-moraes: In this way, places that use Popover will use the click as a default, what changes is that. This change was made precisely because, as the previous implementation did not work on mobile (as shown in the video in which the bug was reported), and the hover is not an expected behavior of a Popover, but of a tooltip. 
Remembering that this was discussed with the team before it was implemented, but we can open it up again for discussions, no problem :)

packages/yoga/src/Popover/web/Popover.jsx
flavia-moraes: This implementation will not work, unless you use an arrow function before it, something like this
`() => toggleOpenPopover()`
The updated code is:
` onKeyDown={event =>
            event.key === 'Enter' ? toggleOpenPopover : undefined
          }`

packages/yoga/src/Popover/web/Popover.jsx
MicaelRodrigues: Let's apply the changes discussed offline and everything will work well

matheushdsbr (approved):

MicaelRodrigues (approved):

PR #682, opened 2023-09-04
feat: prepend element to Input

christian-benseler-gympass:
## Description 📄

Prepend any element to the Input element, inside the bordered area:
Captura de ecrã 2023-09-04, às 10 06 46

Added 2 props to the Input component:

- prependElement: which accepts a node (like the rightIcon prop) 
- prependElementWidth: the width of the element passed as prependElement

We proposed the name of the prop do use _prepend_ because other frameworks use this nomenclature, but it can be left to  keep consistency with the already existed _rightIcon_ prop; we used _Element_ to indicate that this prop accepts any element.

The prependElementWidth is used to add a text indent to the Field component, so the text will be indented (without this prop, the text would be placed above the prepended element). Also, there is some ways to avoid this prop, like calculate the size of the rendered element using the ResizeObserver or with the useLayoutEffect hook. But we thought this approach would be simpler and more effective, because the prepended element is well known by the parent component and it should now its size (and we avoid extra calculation).
Captura de ecrã 2023-09-04, às 10 05 54
Also, the element is always vertically centered.

Code example:
 
Captura de ecrã 2023-09-04, às 10 27 29

## Platforms 📲

- [X] Web
- [ ] Mobile

## Type of change 🔍

- [X] New feature (non-breaking change which adds functionality)

## Checklist: 🔍

- [X] My code follows the contribution guide of this project [Contributing Guide](https://github.com/Gympass/yoga/blob/master/CONTRIBUTING.md)
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] My changes generate no new warnings
- [X] Any dependent changes have been merged and published in downstream modules
- [X] I have checked my code and corrected any misspellings

### DIFF
--- a/packages/yoga/src/Input/web/Input.jsx
+++ b/packages/yoga/src/Input/web/Input.jsx
@@ -105,8 +105,8 @@ const Input = React.forwardRef(
       rightIcon,
       a11yId,
       includeAriaAttributes,
-      prependElement,
-      prependElementWidth,
+      leftElement,
+      leftElementWidth,
       ...props
     },
     ref,
@@ -153,9 +153,7 @@ const Input = React.forwardRef(
           style={style}
           value={value}
         >
-          {prependElement && (
-            <PreprendElement>{prependElement}</PreprendElement>
-          )}
+          {leftElement && <PreprendElement>{leftElement}</PreprendElement>}
           {!children ? (
             <Field
               {...props}
@@ -172,9 +170,7 @@ const Input = React.forwardRef(
               value={value}
               onChange={onChange}
               {...a11yFieldProps}
-              style={
-                prependElementWidth ? { textIndent: prependElementWidth } : {}
-              }
+              style={leftElementWidth ? { textIndent: leftElementWidth } : {}}
             />
           ) : (
             children
@@ -261,10 +257,10 @@ Input.propTypes = {
   /** useful for components that extend the Input component and have their own ARIA attributes implementation (e.g. Dropdown) */
   includeAriaAttributes: bool,
 
-  /** prepend an element */
-  prependElement: node,
-  /** the width of the prepended element */
-  prependElementWidth: string,
+  /** element on the left */
+  leftElement: node,
+  /** the width of the element on the left */
+  leftElementWidth: string,
 };
 
 Input.defaultProps = {
@@ -287,8 +283,8 @@ Input.defaultProps = {
   rightIcon: undefined,
   a11yId: undefined,
   includeAriaAttributes: true,
-  prependElement: undefined,
-  prependElementWidth: undefined,
+  leftElement: undefined,
+  leftElementWidth: undefined,
 };
 
 export default Input;
--- a/packages/yoga/src/Input/web/Input.jsx
+++ b/packages/yoga/src/Input/web/Input.jsx
@@ -78,7 +78,7 @@ const IconWrapper = styled.div`
   `}
 `;
 
-const PreprendElement = styled.div`
+const LeftElementWrapper = styled.div`
   position: absolute;
   left: 8px;
   top: 50%;
@@ -153,7 +153,9 @@ const Input = React.forwardRef(
           style={style}
           value={value}
         >
-          {leftElement && <PreprendElement>{leftElement}</PreprendElement>}
+          {leftElement && (
+            <LeftElementWrapper>{leftElement}</LeftElementWrapper>
+          )}
           {!children ? (
             <Field
               {...props}
--- a/packages/yoga/src/Input/web/Input.jsx
+++ b/packages/yoga/src/Input/web/Input.jsx
@@ -78,6 +78,13 @@ const IconWrapper = styled.div`
   `}
 `;
 
+const PreprendElement = styled.div`
+  position: absolute;
+  left: 8px;
+  top: 50%;
+  transform: translateY(-50%);
+`;
+
 const Input = React.forwardRef(
   (
     {
@@ -98,6 +105,8 @@ const Input = React.forwardRef(
       rightIcon,
       a11yId,
       includeAriaAttributes,
+      prependElement,
+      prependElementWidth,
       ...props
     },
     ref,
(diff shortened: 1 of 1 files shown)


packages/yoga/src/Input/web/Input.jsx
```diff
@@ -98,6 +105,8 @@ const Input = React.forwardRef(
       rightIcon,
       a11yId,
       includeAriaAttributes,
+      prependElement,
+      prependElementWidth,
```
MicaelRodrigues, line 109: This prop shouldn't be necessary. The size of the prependElement should be defined by it's own component. 
Do you agree?

packages/yoga/src/Input/web/Input.jsx
```diff
@@ -160,6 +172,9 @@ const Input = React.forwardRef(
               value={value}
               onChange={onChange}
               {...a11yFieldProps}
+              style={
+                prependElementWidth ? { textIndent: prependElementWidth } : {}
```
MicaelRodrigues, line 176: This should not be needed if the input uses `flexbox` or `grid` code. Just set a `gap` between the elements would work :)

packages/yoga/src/Input/web/Input.jsx
```diff
@@ -98,6 +105,8 @@ const Input = React.forwardRef(
       rightIcon,
       a11yId,
       includeAriaAttributes,
+      prependElement,
```
MicaelRodrigues, line 108: WDYT of naming this `leftElement` or `leftContent`?

packages/yoga/src/Input/web/Input.jsx
christian-benseler-gympass: The _prependElementWidth_ is needed to add a text-indent to the styled-component's input, so the text start next to the element. Without the prop, it would be necessary to add some calculation inside the existing Input.tsx to detect the width from the rendered element after the DOM render.

wellingtoncosta-gympass (approved):

packages/yoga/src/Input/web/Input.jsx
christian-benseler-gympass: I don't get how this would work @MicaelRodrigues 
The element to be rendered is inside another one, the PreprendElement, that has an absolute position (the same approach used for the right icon).

image
image

That's the reason the Field has an indent.

To use the flexbox, the parent component (Fieldset) would need to be changed but this can lead to some unhandled scenarios.

packages/yoga/src/Input/web/Input.jsx
christian-benseler-gympass: Makes sense, so it will use the simular nomenclature from the right positioned icon.

packages/yoga/src/Input/web/Input.jsx
MicaelRodrigues: > To use the flexbox, the parent component (Fieldset) would need to be changed but this can lead to some unhandled scenarios.

I'd prefer to change the code than to introduce a prop the forces the user to define a width of an element to indent the text... It's not very responsive. After all, it's just css. 
> I don't get how this would work @MicaelRodrigues
I think `PhoneInput` has an implementation like the one i'm suggesting.  But, I can try to find a time to help you when i got back to work next week :)

christian-benseler-gympass: My main concern in changing the code to use flexbox is that everything is absolute positioned inside the Input component.
The input, the label, in the DOM:
image

the styles:
image
from the input

image
from the label

I guess those elements are absolute to force their positions inside the parent one, the Fieldset. I don't see an _easy_ way to refactor this only to replicate a similar UI from the right positioned icon.


About the width, I guess the UI won't break, because the current Input behavior doesn't change. If there is no leftElement prop, nothing changes. And if a parent component passes this prop, it is its responsibility to know the width of the element it is passing to a child component. In other words, the parent component must know if it is rendering an icon width 20ox of width or a credit card flag with 48px.

christian-benseler-gympass:
@MicaelRodrigues I made a refactor in the implementation and that prop leftElementWidth is not required anymore.
Now, when there is a leftElement prop defined, the component uses the useLayoutEffect hook to calculate the width of the element. The hook will only be triggered when the leftElement is defined, and then it avoids to be recalculated every time.

ezgif-3-22b8ccb5e2

MicaelRodrigues:
> @MicaelRodrigues I made a refactor in the implementation and that prop leftElementWidth is not required anymore. Now, when there is a leftElement prop defined, the component uses the useLayoutEffect hook to calculate the width of the element. The hook will only be triggered when the leftElement is defined, and then it avoids to be recalculated every time.
> 
> ezgif-3-22b8ccb5e2 [ ezgif-3-22b8ccb5e2 ](https://user-images.githubusercontent.com/135242379/268598454-4bf94551-455e-4c47-ac3f-0961ed0665c6.gif) [ ](https://user-images.githubusercontent.com/135242379/268598454-4bf94551-455e-4c47-ac3f-0961ed0665c6.gif)

Excellent! :) I prefer that approach, because when there is an opportunity when can refactor to flex/grid without "deprecating" props

packages/yoga/src/Input/web/Input.jsx
```diff
@@ -78,6 +78,13 @@ const IconWrapper = styled.div`
   `}
 `;
 
+const LeftElementWrapper = styled.div`
+  position: absolute;
+  left: 8px;
```
MicaelRodrigues, line 83: Check if it'd better to use yoga spacing props

MicaelRodrigues (approved):

packages/yoga/src/Input/web/Input.jsx
christian-benseler-gympass: Thanks for the tip. Refactored to use the _spacing_ prop with the _xxsmall_ variation.

luanlorenzo (approved):

joaopaulonsoares-gympass (approved):

matheushdsbr (approved):

PR #705, opened 2023-10-17
🚀 feat(button): add prop loading

matheushdsbr:

## Description 📄

This is a new prop of button, now we can use loading on buttons! ✨

We can use it when we are going to perform some type of loading or request, and it will take a while to finish. That's when the loading comes into action.

Exemple:
`<Button isLoading>Find an activity</Button>`
`<Button isLoading={false}>Find an activity</Button>`

## Platforms 📲

- [x] Web
- [ ] Mobile

## Type of change 🔍

- [x] Unit Test
- [x] Snapshot Test

## Checklist: 🔍

- [x] Layout matches design prototype: [FIGMA](https://www.figma.com/file/xz8x4jA5gwPSSp0StCdB7V/Hackaton---Loaders?type=design&mode=design&t=OmQ1ZMWv8pa8TzVB-0)

### Default Button
https://github.com/gympass/yoga/assets/32910717/f3e00bca-ee74-4d6b-99f0-286795149cf9

### Icon Button
https://github.com/gympass/yoga/assets/32910717/55813f47-74ba-4d92-bb45-1d5d1c1832ec

### Outline Button
https://github.com/gympass/yoga/assets/32910717/ab8423f8-0cf1-4033-a540-b96715e9d6cd


frgiovanna:
Why not make the `isLoading` an optional prop for the already existent buttons? 

(On one of our projects have an abstraction that does what I am suggesting, check [here](https://github.com/gympass/olympus/blob/master/src/shared/components/Buttons/Button.tsx))

joaopaulonsoares-gympass:
I agree with @frgiovanna , have loading as props would be better than a new component.

caiotracera:
As said by @frgiovanna and @joaopaulonsoares-gympass, I really think that we should change this, so the loading state can be a prop instead of a whole new component.

packages/doc/content/components/components/button/default-web.mdx
```diff
@@ -14,6 +14,14 @@ You can use all differente compounds like `<Button.Outline />` and `<Button.Text
 </Box>
 ```
 
+#### Inverted
```
frgiovanna, line 17: 🏅

packages/doc/content/components/components/button/loading.mdx
frgiovanna, line 1: Do we still need this?

packages/yoga/src/Button/web/Link.jsx
```diff
@@ -68,9 +68,14 @@ const Link = styled(Button)`
   }}
 `;
 
-const ButtonLink = ({ disabled, ...props }) => (
-  <Link {...props} disabled={disabled} aria-disabled={disabled} />
-);
+// eslint-disable-next-line react/prop-types
+const ButtonLink = forwardRef(({ isLoading, ...rest }, ref) => {
```
frgiovanna, line 72: Where are we using this destructured `isLoading`?

packages/yoga/src/Button/web/Link.jsx
```diff
@@ -68,9 +68,14 @@ const Link = styled(Button)`
...
+  const { disabled } = rest;
+
+  return (
+    <Link {...rest} disabled={disabled} aria-disabled={disabled} ref={ref} />
+  );
+});
 
 ButtonLink.propTypes = {
```
frgiovanna, line 80: Shouldn't the isLoading be typed here?

packages/yoga/src/Button/web/Text.jsx
```diff
@@ -81,6 +83,30 @@ const ButtonText = styled(Button)`
...
+    <StyledButton
+      {...rest}
+      secondary={secondary}
+      inverted={inverted}
+      ref={ref}
+    />
+  );
+});
+
+ButtonText.propTypes = {
```
frgiovanna, line 100: Here too about the isLoading prop type

frgiovanna (commented):
Good job so far, please check my comments

packages/doc/content/components/components/button/loading.mdx
matheushdsbr: ops!! I'll remove it!

packages/yoga/src/Button/web/Link.jsx
matheushdsbr: My intention here is to avoid using 'isLoading' when working with 'button.link' and 'button.text.' However, when I include the 'isLoading' types, it also adds this information to the yoga documentation, suggesting that 'isLoading' might be applicable.

frgiovanna: Got it, I didn't notice that these button types wouldn't have a loading state

packages/yoga/src/Button/web/Link.jsx
frgiovanna: Do you think is scalable if you get all of the other props except the `isLoading`? This means every time someone adds a new value for all buttons they will need to add it here too, otherwise it will be ignored. WDYT?

matheushdsbr: Good point 🤔

I'm just concerned that someone might misunderstand the 'button.text' and 'button.link' props and try to use them. By removing this approach, people will still be able to use isLoading, and it will work. However, it wouldn't be advisable to do so because, as far as I know, we don't have any plans for that.

But, yes I agree with you. This approach isn't scalable. It's more on the developer's side to use it correctly.

frgiovanna: No no, I really got your point and I think you're right, I'm just wondering if we can handle it in another way instead of having this eslint comment

frgiovanna: What if you do something like this? 

```JSX
const ButtonLink = forwardRef(({ rest }, ref) => {

const props = Object.fromEntries(
  Object.entries(rest).filter(([key]) => key !== 'isLoading')
);
  
<Link {...props} disabled={props.disabled} aria-disabled={props.disabled} ref={ref} /> 
```

matheushdsbr: Great suggestion!! 🚀

frgiovanna (approved):

packages/yoga/src/Button/web/Button.jsx
```diff
@@ -14,10 +16,22 @@ const Button = forwardRef(
...
+    const SpinnerContainer = styled.div`
+      ${() =>
+        isLoading &&
+        css`
+          position: absolute;
+          top: 50%;
+          left: 50%;
+          transform: translate(-50%, -50%);
+        `}
+    `;
```
MicaelRodrigues, line 33: A component should not be created inside render. It's a bad practise. Here is why:

https://www.youtube.com/watch?app=desktop&v=2sAdzy90GtE&feature=youtu.be

MicaelRodrigues (changes requested):

packages/yoga/src/Button/web/Button.jsx
matheushdsbr: Done! Thanks for your review!! 🚀

MicaelRodrigues: @matheushdsbr `    const SpinnerContainer = styled.div` is still inside the `Button` function component. It should be created outside

matheushdsbr: Done! o/

packages/yoga/src/Spinner/web/Spinner.jsx
```diff
@@ -60,6 +75,7 @@ const Spinner = React.forwardRef(({ size, color, theme }, ref) => {
       color={get(theme.yoga.colors, color, color)}
       size={get(theme.yoga.spacing, size, size)}
       ref={ref}
+      aria-label="loading-icon"
```
flavia-moraes, line 78: Isso também já viria pré definido?

matheushdsbr: sim, adicionei para realizar testes lá no button

MicaelRodrigues (approved):

PR #733, opened 2024-02-19
🚀[RTN-858] - feat(Heading): New heading component style

agnezi:

## Description 📄

We are adding the new `Heading` component style.

## Platforms 📲

- [x] Web
- [ ] Mobile

## Type of change 🔍

- [ ] Unit Test
- [ ] Snapshot Test

## Checklist: 🔍

- [x] Layout matches design prototype: [FIGMA](https://www.figma.com/file/Hsrgmg4oLjJxYcJHaFeuJP/%F0%9F%8E%A8-Yoga-3?type=design&node-id=1079-40&mode=design&t=1xjUGbdFLofIUgwh-0)

## Screenshots 📸

| Before | After - V2 | After - V3 |
| ------ | ----- | ----- |
| Screenshot 2024-02-19 at 10 21 09 | Screenshot 2024-02-19 at 10 18 38 | Screenshot 2024-02-19 at 10 23 43 |

### DIFF
--- a/packages/yoga/src/Heading/web/BackButton.jsx
+++ b/packages/yoga/src/Heading/web/BackButton.jsx
@@ -3,6 +3,7 @@ import styled from 'styled-components';
 import PropTypes from 'prop-types';
 import { ArrowLeft } from '@gympass/yoga-icons';
 import Button from '../../Button';
+import Box from '../../Box';
 
 const ButtonIcon = styled(Button.Icon)`
   ${({
@@ -20,7 +21,9 @@ const ButtonIcon = styled(Button.Icon)`
 `;
 
 const BackButton = ({ onClick, ...props }) => (
-  <ButtonIcon {...props} icon={ArrowLeft} inverted onClick={onClick} />
+  <Box padding="xxxsmall" onClick={onClick} role="button" tabIndex={0}>
+    <ButtonIcon tabIndex={-1} icon={ArrowLeft} inverted secondary {...props} />
+  </Box>
 );
 
 BackButton.propTypes = {
--- a/packages/yoga/src/Heading/web/BackButton.jsx
+++ b/packages/yoga/src/Heading/web/BackButton.jsx
@@ -4,9 +4,11 @@ import PropTypes from 'prop-types';
 import { ArrowLeft } from '@gympass/yoga-icons';
 import Button from '../../Button';
 import Box from '../../Box';
+import Icon from '../../Icon';
 
-const ButtonIcon = styled(Button.Icon)`
+const ButtonIcon = styled(Button)`
   ${({
+    secondary,
     theme: {
       yoga: {
         components: { heading },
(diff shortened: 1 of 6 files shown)


packages/yoga/src/Heading/web/BackButton.jsx
```diff
@@ -1,26 +1,78 @@
...
-import styled from 'styled-components';
+import styled, { css } from 'styled-components';
 import PropTypes from 'prop-types';
 import { ArrowLeft } from '@gympass/yoga-icons';
 import Button from '../../Button';
+import Box from '../../Box';
+import Icon from '../../Icon';
 
-const ButtonIcon = styled(Button.Icon)`
+const ButtonIcon = styled(Button)`
```
leonardo-gympass, line 9: There are other components that will have the same change in `Button.Icon`, such as `Drawer`, shouldn't this style be a new variant of `Button.Icon` so it can be reused?

FYI @eduardobreno @rafaelcoleta

agnezi: Maybe, only implementing the possibility to pass a `children` to `Button.Icon`, and using `icon` from props or `children` will be sufficient here. What do you think? 

Because most of the styles I added for the `buttons` on the `Heading` component, were just to make the `button` have some style like the `Button.Icon`.

leonardo-gympass: I think it will make it easier to reuse if this is part of the `Button.Icon` spec, the change looks like one of the variants in the "Usage" section, with a change in colors.
https://gympass.github.io/yoga/components/button/icon#usage

tcK1: Hey team! This is a good suggestion, and I'm on board that this is something that we should do. But maybe this is not the moment.

I think that in this first moment, we can introduce this duplication as a technical debt and refactor this in the future.

Our main concern is not the code, but our documentation and Figma (for designers). We would create a new variation to the `Button` component, and this needs to have the necessary attention.

Does this make sense?

leonardo-gympass: That's fine, but in the future we will have rework in at least 3 different usage scenarios.

leonardo-gympass, line 9: What do you think about the approach bellow where i'm using the `secondary` and `inverted` props from `Button.Icon`? This way we only need to override some `Button.Icon` styles.

https://github.com/gympass/yoga/pull/735/files#diff-d4cfcfeaeb3e6306f1915918e710fcae9615d3fa12d829e229ee1684962e33d8

agnezi: hmm... maybe in my case it won't work... because I have a visible area and a clickable area with different sizes, this is why I'm adding the `Icon` inside the `Button` and working on the sizes.

[You can see it here](https://www.figma.com/file/Hsrgmg4oLjJxYcJHaFeuJP/%F0%9F%8E%A8-Yoga-3?type=design&node-id=2203-354&mode=design&t=Ewh5gxYSMxZyFkmN-4)

leonardo-gympass: I hadn't noticed this click area, thanks!

leonardo-gympass (approved):

tcK1 (approved):

evilamaior (approved):

matheushdsbr (approved):

PR #745, opened 2024-03-01
🚀 [EUEG-3173] - feat(Heading): allow to override the background color

leonardo-gympass:
[JIRA Issue](https://gympass.atlassian.net/browse/EUEG-3173)

## Description 📄

Allow to override the background color from `Header` component.

```jsx
<Heading bg="yoga">
  <Heading.Title>Gympass</Heading.Title>
</Heading>
```

## Platforms 📲

- [x] Web
- [ ] Mobile

## Type of change 🔍

- [ ] Unit Test
- [x] Snapshot Test

## Checklist: 🔍

## Screenshots 📸
localhost_8000_components_heading(iPhone SE)

### DIFF
--- a/packages/yoga/src/Heading/web/StyledHeading.jsx
+++ b/packages/yoga/src/Heading/web/StyledHeading.jsx
@@ -1,6 +1,11 @@
 import styled, { css } from 'styled-components';
+import Box from '../../Box';
 
-const StyledHeading = styled.header`
+const StyledHeading = styled(Box).attrs(props => ({
+  as: 'header',
+  bg: props.theme.yoga.components.heading,
+  ...props,
+}))`
   ${({
     noPadding,
     theme: {
@@ -10,7 +15,6 @@ const StyledHeading = styled.header`
     },
   }) => {
     return css`
-      background: ${heading.background};
       padding: ${heading.padding.vertical}px ${heading.padding.horizontal}px;
       min-height: ${heading.height}px;
       width: 100%;
--- a/packages/doc/content/components/components/heading/index.mdx
+++ b/packages/doc/content/components/components/heading/index.mdx
@@ -4,7 +4,6 @@ metaTitle: 'Heading'
 metaDescription: 'Heading Component'
 ---
 
-
 # Heading
 
 ### Reference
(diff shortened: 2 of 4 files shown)


aly-lopes (approved):

eduardobreno (approved):

ruanblimaig (approved):

packages/yoga/src/Heading/web/StyledHeading.jsx
```diff
@@ -1,6 +1,11 @@
 import styled, { css } from 'styled-components';
+import Box from '../../Box';
 
-const StyledHeading = styled.header`
+const StyledHeading = styled(Box).attrs(props => ({
+  as: 'header',
+  bg: props.theme.yoga.components.heading,
+  ...props,
+}))`
```
ericcleao, line 8: The Box component should not be used inside the Yoga by definition, if we add it here ti will bring to the component some extra props as `margins`, for instance.
Could we add the background color support for the yoga system approach, WDYT?

```suggestion
import { backgroundColor } from '@gympass/yoga-system';

const StyledHeading = styled.header`
  ${backgroundColor}
```

leonardo-gympass: It makes sense @ericcleao, I made the change you suggested. Thanks!

matheushdsbr (approved):

tcK1 (approved):

packages/yoga/src/Heading/web/Heading.jsx
```diff
@@ -43,11 +43,13 @@ Heading.propTypes = {
     PropTypes.node,
   ]),
   noPadding: bool,
+  bg: PropTypes.string,
```
MicaelRodrigues, line 46: A more meaningful name:
```suggestion
  backgroundColor: PropTypes.string,
```
Also, maybe we can start using [transient props](https://styled-components.com/docs/api#transient-props) naming so these "style" props don't go to the DOM.

```suggestion
  $backgroundColor: PropTypes.string,
```

evilamaior (approved):

packages/yoga/src/Heading/web/Heading.jsx
leonardo-gympass: This name is part of the styled system, we can also use the `backgroundColor` and `bgColor` attributes to define the background color although they are not listed in the prop-types. When we start using typescript it will be possible to list all these props automatically.
As these props are part of the styled system, they are not being injected into the DOM.

PR #823, opened 2024-05-27
🐞fix[tokens/system]: Adjust export breakpoints types

matheushdsbr:

## Description 📄

- We have an error because BreakpointsKeys and BreakpointsTypes are not being found.

- This is because it looks for these types in `esm` or `cjs` and doesn't find them, they only appear in `typings`. Since it can't find them, it breaks.

- So one solution to this problem is that since BreakpointsKeys and BreakpointsTypes are only used in the helpers types, we can use an import with the direct path instead of using the package alias

Exemple:

```
import { BreakpointsKey, BreakpointsType } from '../../tokens/src/global/breakpoints';
```

instead of:

```
import { BreakpointsKey, BreakpointsType } from '@gympass/yoga-tokens';. 
```

- This way the build will be executed and the types will appear at development time and also in the `typings` folder generated by the build. The `esm` and `cjs` packages will build normally.

## Platforms 📲

- [x] Web
- [x] Mobile

## Type of change 🔍

- [ ] Unit Test
- [ ] Snapshot Test

## Checklist: 🔍

### DIFF
--- a/packages/helpers/src/hide.ts
+++ b/packages/helpers/src/hide.ts
@@ -1,4 +1,5 @@
-import tokens, { BreakpointsType }  from '@gympass/yoga-tokens';
+import tokens  from '@gympass/yoga-tokens';
+import { BreakpointsType } from '../../tokens/src/global/breakpoints';
 
 const { breakpoints } = tokens;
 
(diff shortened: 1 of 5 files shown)


aly-lopes (approved):
❓ What if instead we exported these types by adding a new build step running TS with `--emitDeclarationOnly` https://www.typescriptlang.org/tsconfig/#emitDeclarationOnly?

That way the types could be accessible for whoever uses this package.

PR #849, opened 2024-11-18
feat(Dialog): improve accessibility

leonardo-gympass:

## Description 📄

Add `react-focus-lock` library to `@gympass/yoga` package, with this library the following features were added to `Dialog`, `BottomSheet` and `Drawer` components:
- Create a focus trap inside the `Dialog` when it is opened, that way the user can't move the focus away from the `Dialog`.
- Focus on the last active element when the Dialog is closed.
- Focus on the first programmatic focusable element when the Dialog is opened.

Add props `role="dialog"` and `aria-modal=true` to `Dialog` component.
Add `closeLabel` prop to be used as the `aria-label` from close button.

## Platforms 📲

- [x] Web
- [ ] Mobile

## Type of change 🔍

- [ ] Unit Test
- [ ] Snapshot Test

## Checklist: 🔍

## Screenshots 📸

| Before | After |
| ------ | ----- |
|        |       |

### DIFF
--- a/packages/yoga/src/Dialog/web/Dialog.jsx
+++ b/packages/yoga/src/Dialog/web/Dialog.jsx
@@ -1,11 +1,13 @@
 import React, { useEffect, useCallback } from 'react';
 import { createPortal } from 'react-dom';
+import FocusLock from 'react-focus-lock';
 import styled, { css } from 'styled-components';
 import { func, bool, node, number, string } from 'prop-types';
 
 import { Close } from '@gympass/yoga-icons';
 import { usePortal, useCombinedRefs } from '../../hooks';
 import { Button, Card, Box } from '../..';
+import { focusOnFirstProgrammaticFocusableElement } from './utils';
 
 export const StyledDialog = styled(Card)`
   ${({
@@ -119,7 +121,16 @@ const Dialog = React.forwardRef(
           ref={dialogRef}
           zIndex={zIndex}
         >
-          <StyledDialog onClose={onClose} {...props}>
+          <FocusLock
+            as={StyledDialog}
+            role="dialog"
+            aria-modal
+            returnFocus
+            disabled={!isOpen}
+            onClose={onClose}
+            onActivation={focusOnFirstProgrammaticFocusableElement}
+            {...props}
+          >
             {isCloseButtonVisible && (
               <Box d="flex" justifyContent="flex-end" w="100%">
                 <CloseButton
@@ -131,7 +142,7 @@ const Dialog = React.forwardRef(
               </Box>
             )}
             {children}
-          </StyledDialog>
+          </FocusLock>
         </Overlay>,
         dialogElement,
       )
--- a/packages/yoga/src/Dialog/web/Dialog.jsx
+++ b/packages/yoga/src/Dialog/web/Dialog.jsx
@@ -95,6 +95,11 @@ const Dialog = React.forwardRef(
     const dialogRef = useCombinedRefs(forwardedRef);
     const dialogElement = usePortal(dialogId ?? 'dialog');
     const isCloseButtonVisible = onClose && !hideCloseButton;
+    const lockProps = {
+      role: 'dialog',
+      'aria-modal': true,
+      onClose,
+    };
 
     const closeDialog = useCallback(
       e => {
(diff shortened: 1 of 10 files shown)


angelo-gympass (approved):

packages/yoga/src/Dialog/web/utils.js
```diff
@@ -0,0 +1,5 @@
+export function focusOnFirstProgrammaticFocusableElement(container) {
+  const element = container.querySelector('[tabindex="-1"]');
+
+  element?.focus();
+}
```
nataliosilveiraext, line 5: Why are you focusing on items that have been removed from navigation?

If an element is not naturally focusable, we use tabindex="0"

As we are using react, I believe that useRef is more suitable for focusing on elements.

Use .focus() only when necessary.
Apply tabindex="0" to non-focusable but relevant elements.
Ensure focus cycling within components such as modals.
Add visual feedback on focus when necessary.
Work with accessibility attributes to guide assistive technologies.
These practices ensure an inclusive and consistent experience for all users.

packages/yoga/src/Dialog/web/Header.jsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import Text from '../../Text';
 
 const Header = props => (
-  <Text.H4 bold as="header" ta="center" mb="large" {...props} />
+  <Text.H4 bold as="h2" ta="center" mb="large" tabIndex={-1} {...props} />
```
nataliosilveiraext, line 5: I don't think tabindex="-1" is necessary in this situation

packages/yoga/src/Dialog/web/Dialog.jsx
```diff
@@ -119,10 +137,18 @@ const Dialog = React.forwardRef(
...
           zIndex={zIndex}
         >
-          <StyledDialog onClose={onClose} {...props}>
+          <FocusLock
+            as={StyledDialog}
+            lockProps={lockProps}
+            returnFocus
+            disabled={!isOpen}
+            className={className}
+            onActivation={focusOnFirstProgrammaticFocusableElement}
```
nataliosilveiraext, line 146: be careful when focusing with tabindex="-1", it is not a good practice

nataliosilveiraext (changes requested):

packages/yoga/src/Dialog/web/utils.js
leonardo-gympass: This function is called by the `onActivation` prop from `FocusLock`, in this case the container will be the `FocusLock` itself, according to Figma's specs in most scenarios the initial focus will be on the title.

These adjustments were made following the w3c practices guide, the only attribute I didn't add was `aria-labelledby` because of the `Drawer` scenario, but since the initial focus will always be on the title I don't think this will be a problem.
https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/

The use of `tabindex=-1` is for scenarios where a non-interactive element is focused programmatically.
https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/

packages/yoga/src/Dialog/web/Header.jsx
leonardo-gympass: The recommendation for the initial focus of a Dialog varies according to the context. In our case, we are always focusing on the first element of the dialog, which would be the title in most cases.

To add the initial focus to a non-interactive element, it is recommended to use tabindex="-1".

https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/
Screenshot 2024-11-19 at 08 21 20

lucascapellini-wellhub (approved):

nataliosilveiraext (approved):

diogoverde (approved):

packages/yoga/src/Dialog/web/Dialog.jsx
```diff
@@ -119,10 +137,18 @@ const Dialog = React.forwardRef(
           ref={dialogRef}
           zIndex={zIndex}
         >
-          <StyledDialog onClose={onClose} {...props}>
+          <FocusLock
+            as={StyledDialog}
+            lockProps={lockProps}
+            returnFocus
+            disabled={!isOpen}
+            className={className}
```
flavia-moraes, line 145: What is the use of className here? Can this allow style customization?

PR #866, opened 2025-01-24
fix: remove default props from yoga/web

lpolon:
[JIRA Issue](https://gympass.atlassian.net/browse/)

## Description 📄

This PR removes React `defaultProps` in favor of js default parameters

React 19 removed the PropTypes and defaultProps for functions ([link](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#removed-deprecated-react-apis)).

relevant links:
- https://react.dev/blog/2024/04/25/react-19-upgrade-guide
- https://stackoverflow.com/questions/78931633/warning-page-support-for-defaultprops-will-be-removed-from-function-components

related issues:
https://github.com/gympass/yoga/issues/857#issue-2757561466

[Enter the description of your changes]

## Platforms 📲

- [x] Web
- [ ] Mobile

## Type of change 🔍

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How Has This Been Tested? 🧪

[Enter the tips to test this PR]

- [ ] Unit Test
- [x] Snapshot Test

## Checklist: 🔍

- [x] My code follows the contribution guide of this project [Contributing Guide](https://github.com/Gympass/yoga/blob/master/CONTRIBUTING.md)
- [ ] Layout matches design prototype: [FIGMA](https://figma.com/file/YOUR_LINK)
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes -> WIP
- [ ] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

PR #867, opened 2025-01-27
fix(ButtonIcon): remove unnecessary `aria-disabled` from component

lpolon:
## Description 📄

Previously, when the disabled prop was omitted, the attribute `disabled` would resolve to `false` due to `isLoading` default value, but `aria-disabled` would be omitted due to passing `undefined` value to a styled component removes the attribute:

```jsx
const ButtonIcon = forwardRef(
  (
    {
      // ...
      disabled, // defaults to undefined
      isLoading, // defaults to false
      ...props
    },
    ref,
  ) => {
    return (
      <IconStyled
        {...props}
        ref={ref}
        small={small}
        disabled={disabled || isLoading} // rendered as disabled='false'
        aria-disabled={disabled} // ignored
        isLoading={isLoading}
      >
);
```
It could be fixed, however it is unnecessary. The disable attribute in buttons is enough ([stack overflow](https://stackoverflow.com/questions/38059140/when-to-use-the-disabled-attribute-vs-the-aria-disabled-attribute-for-html-eleme)). adding `aria-disabled` does not make any difference in the screen reader experience:

https://github.com/user-attachments/assets/b4f37c61-d238-4d18-851e-ec70345a64c2

Moreover, the current implementation disallow extensibility of aria-disabled through custom controls, for instance, The `Button.Icon` in the Menu test is being controlled by a Radix component that might implement a different disabled behavior that might use aria-disabled (I am saying "might" because it does disable the button, but using data-disabled and other a11y-attributes. I don't have a proper use case.)

I stumbled into it while removing defaultProps:
related PR: https://github.com/gympass/yoga/pull/866

## Platforms 📲

- [x] Web
- [ ] Mobile

## Type of change 🔍

- [ ] Unit Test
- [x] Snapshot Test

## Checklist: 🔍

### DIFF
--- a/packages/yoga/src/Button/web/Icon.jsx
+++ b/packages/yoga/src/Button/web/Icon.jsx
@@ -51,7 +51,6 @@ const ButtonIcon = forwardRef(
         ref={ref}
         small={small}
         disabled={disabled || isLoading}
-        aria-disabled={disabled}
         isLoading={isLoading}
       >
         {isLoading ? (
--- a/packages/yoga/src/Button/web/__snapshots__/Button.test.jsx.snap
+++ b/packages/yoga/src/Button/web/__snapshots__/Button.test.jsx.snap
@@ -558,7 +558,6 @@ exports[`<Button /> Snapshots primary buttons With inverted prop With disabled p
 
 <div>
   <button
-    aria-disabled="true"
     class="c0"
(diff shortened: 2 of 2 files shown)


wellhub-gsevla (approved):
Nice catch!

